PR #74, opened 2015-01-30
Do not reinterpret std::size_t on windows

springmeyer:
This fixes:

```
libosmium\include\osmium/io/detail/read_write.hpp(46): error C2371: 'ssize_t': redefinition
```

refs https://github.com/osmcode/node-osmium/issues/32

### DIFF
--- a/include/osmium/io/detail/read_write.hpp
+++ b/include/osmium/io/detail/read_write.hpp
@@ -39,6 +39,12 @@ DEALINGS IN THE SOFTWARE.
 #include <string>
 #include <system_error>
 
+#ifndef _MSC_VER
+# include <unistd.h>
+#else
+# include <io.h>
+#endif
+
 #include <osmium/io/overwrite.hpp>
 
 namespace osmium {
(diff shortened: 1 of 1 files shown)


joto:
I am not sure this is the right approach. There are two other variables (`size` and `offset`) involved and their types have to fit, too. Has to work on all OSes and on 32bit/64bit. (Although 32 bit build is half-broken currently anyway.) I have to think about this some more and do more tests.

Now that the appveyor stuff seems to be working, we have at least the option of checking on all systems.

springmeyer:
Okay, let me know what you think is the right approach. But it is clear that `int != std::ssize_t` on windows always.

joto:
I looked into this more closely and we indeed have a larger problem here. Even with your patch there is a warning from MSVC: `conversion from 'size_t' to 'unsigned int'` which is about the argument to `::write()`. It seems the `count` argument to `write()` is an `unsigned int` which is not enough for files > 4GB. But there might be cases where we want to write more than 4GB of data which will then not work on Windows. The function has to be rewritten to take this into account and write the data in chunks. There might be other parts of the library that are affected by similar problems, too.

PR #79, opened 2015-02-05
add override keyword ...

DennisOSRM:
...to close(), overridden function in include/osmium/io/detail/input_format.hpp


DennisOSRM:
thx, will bump used version in osrm-backend.

PR #95, opened 2015-06-13
Allow extra compiler options to be specified

tomhughes:


joto:
Any reason why you can't use `cmake -DCMAKE_CXX_FLAGS=-Foo` ?

Also the command `add_compile_options` is only available in CMake 3 and I'd prefer if Libosmium works with older CMake.

tomhughes:
Well I tried about a bazillion things before I added that patch, but if that is the magic define then fine - will try it and let you know ;-)

tomhughes:
Yes, that does seem to work - they go before other options, so the -O3 that you add probably wins over the -O2 from the Fedora standard options which is probably technically wrong from a Fedora packaging standpoint but other than that it should work.

PR #96, opened 2015-06-17
Cleaning up the "osmium_count.cpp" example.

r-barnes:
Just a few minor changes to clean up the code's formatting and use of end-lines.


joto:
I don't use `std::endl` on purpose, because it flushes the buffers which leads to bad performance. Not a big deal in this case, but it is the wrong habit to get into. Compilation instructions should probably go to the wiki if the existing instructions are not enough.

PR #108, opened 2015-08-20
GDAL 2 Support

### LINKED ISSUE #107: Add gdal2 support
Make libosmium work with gdal/ogr 1 and 2.

This is currently waiting for gdal 2 support in Debian.

tomhughes:
This is more of an RFC than anything you actually want to merge just yet...

The first commit is fine, and compiled with with both GDAL 1.9 and GDAL 2.0 but the second one break 1.9 due to API incompatibilities.

What I'd like is your thoughts on the best way to support both - should we create some sort of wrapper to wrap the differences in drivers/datasources? and should that be internal for the tests or something that is part of the released libosmium?

In any case this will hopefully resolve #107 and is needed for Fedora 23 which is due out in a few months.

### DIFF
--- a/include/osmium/area/problem_reporter_ogr.hpp
+++ b/include/osmium/area/problem_reporter_ogr.hpp
@@ -59,6 +59,7 @@ DEALINGS IN THE SOFTWARE.
 
 #include <ogr_api.h>
 #include <ogrsf_frmts.h>
+#include <gdal_priv.h>
 
 #ifdef _MSC_VER
 # pragma warning(pop)
@@ -85,8 +86,6 @@ namespace osmium {
         class ProblemReporterOGR : public ProblemReporter {
 
             osmium::geom::OGRFactory<> m_ogr_factory;
-
-            OGRDataSource* m_data_source;
 
             OGRLayer* m_layer_perror;
             OGRLayer* m_layer_lerror;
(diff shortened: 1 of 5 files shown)


joto:
Thanks, Tom. This helps us seeing where the problems are.

We have  basically three different parts of the problem here:
- The code inside libosmium refering to gdal/ogr. There are only two files and they don't do much. It should be possible to either change them so that they work with any gdal version, or to have two versions of those files and you include just one depending on your gdal version. 
- The examples `osmium_toogr` and `osmium_toogr2` need the most changes. I think they should be moved into their own repository/project. As examples they are already too complicated, but on the other hand they are not powerful enough to use as real applications. Most people using libosmium don't need them anyway and in their own project they could develop a life of their own.
- The tests. I am not sure yet how to handle those. The tests are supposed to test the multipolygon code in libosmium, not its gdal code, so they really shouldn't use gdal at all but something simpler. The overview test already doesn't work on Windows. Maybe we can split them up, into the core multipolygon code which will remain in libosmium, and the gdal code which can go into tests in the new repository where the ogr examples will go.

/cc @springmeyer

tomhughes:
So the change to `include/osmium/geom/ogr.hpp` is trivial and works fine with GDAL 1 anyway, but could be left out and clients could do the include - it's basically just as the comment there says already that the include is a convenience.

The other issues internal to the library is `ProblemReporterOGR` and that is also fairly easy to solve I think. Firstly we don't actually need to save `data_source` as a member because we never use it after the constructor is done. So we could just have two constructors, one for each type (both types exist in both GDAL versions) which delegate to a templated private member function to avoid code duplication?

BergWerkGIS:
@joto ping me, when `gdal@2.0` support has landed, that I'll update the Windows binary deps package.
Current one is with `gdal@1.11.2`.

tomhughes:
I've split the `ProblemReporterOGR` out into a separate commit now, which works with GDAL 1 and 2 by just making the first argument to the constructor conditional on the GDAL version.

The delegating to a template approach didn't work because `GDALDataset` doesn't have `CreateLayer` in GDAL 1 so failed on that basis. This approach should be OK because `OGRDataSource` inherits from `GDALDataset` in GDAL 2 so either type will work with only the one constructor.

joto:
As a first step I moved the toogr* examples out of libosmium into their own repository at https://github.com/osmcode/osm-gis-export.

tomhughes:
I've rebased on current master and factored out the differences for the tests into a head that is conditioned on the GDAL version so this should now build for both GDAL 1 and 2.

Tested on Fedora 22 with GDAL 1.11.2 and Fedora Rawhide with GDAL 2.0.0.

joto:
Thanks, Tom. I have a somewhat bigger, but also cleaner solution now. See #107.

PR #110, opened 2015-08-22
Cleanup cmake find module

DerDakon:
Osmium should provide a OsmiumConfig.cmake to allow other CMake projects to integrate it easily. Until someone does it here are some cleanups and improvements to the FindOsmium.cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,10 +131,12 @@ include_directories(${OSMIUM_INCLUDE_DIR})
 
 find_package(Osmium COMPONENTS io gdal geos proj sparsehash)
 
-# The find_package put the directory where it found the libosmium includes as
-# the first item in OSMIUM_INCLUDE_DIRS. We remove it again, because we want
-# to make sure to use our own include directory already set up above.
-list(REMOVE_AT OSMIUM_INCLUDE_DIRS 0)
+# The find_package put the directory where it found the libosmium includes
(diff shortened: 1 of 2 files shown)


joto:
Isn't the `FindOsmium.cmake` exactly that? A way to "allow other CMake projects to integrate it easily."? I have been copying this file over to all my projects using libosmium. What would an `OsmiumConfig.cmake` do differently?

DerDakon:
Well, yes and no. A Find*.cmake module is included in a client project and searches for a library, usually one which is not itself built by CMake. To include FindOsmium.cmake in a client project makes sense, including both the find module and the sources not really, as you do not need to search for anything. There are projects that install both their headers/libraries as well as their find module, so a client project (which does _not_ include the find module, but calls find_package on it) will only work if the package is installed, otherwise CMake will error out because the find module itself is not present.

For projects that are itself built by CMake a *Config.cmake is suggested instead. It avoids all the searching, as one can already know where stuff was installed. It also allows to make use of the CMake language and provide (imported) targets instead of just variables. These targets can carry additional information (as needed compile definitions) which can be consumed by client projects in a way that allows e.g. using those definitions only for the libraries dependend on that target and not others defined in the same project.

joto:
This is getting confusing now. You seem to have changed some commits so I can't find some of our discussion any more regarding the "OSMIUM" vs. "Osmium" changes. I have cherry-picked some of your commits now but the rest needs more discussion. Can we re-start this process somehow?

Issue #112, opened 2015-08-22
CRC checksum test fail on big endian architectures


sebastic:
After including the protozero change (https://github.com/mapbox/protozero/commit/4bd612677dd53d2a4752332a53db8eb87aed67a4) in libosmium (2.3.0-2) the builds of the Debian package now fail on big endian architectures due to CRC checksum mismatches:
```
14: Test command: /«PKGBUILDDIR»/build/test/basic_test_way
14: Test timeout computed to be: 9.99988e+06
14: 
14: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
14: basic_test_way is a Catch v1.2.1 host application.
14: Run with -? for options
14: 
14: -------------------------------------------------------------------------------
14: Build way
14:   way_builder
14: -------------------------------------------------------------------------------
14: /«PKGBUILDDIR»/test/t/basic/test_way.cpp:11
14: ...............................................................................
14: 
14: /«PKGBUILDDIR»/test/t/basic/test_way.cpp:48: FAILED:
14:   REQUIRE( crc32().checksum() == 0x20fe7a30 )
14: with expansion:
14:   2639825081 (0x9d5888b9)
14:   ==
14:   553548336 (0x20fe7a30)
14: 
14: ===============================================================================
14: test cases:  1 |  0 passed | 1 failed
14: assertions: 17 | 16 passed | 1 failed
14: 
```
Build logs: [powerpc](https://buildd.debian.org/status/fetch.php?pkg=libosmium&arch=powerpc&ver=2.3.0-2&stamp=1440280759), [s390x](https://buildd.debian.org/status/fetch.php?pkg=libosmium&arch=s390x&ver=2.3.0-2&stamp=1440281193), (mips is still building at time of writing)

This is pretty much the same problem as: https://github.com/osmcode/osmium-tool/pull/13

joto:
This should be fixed in f081942.

sebastic:
This was actually fixed in 95a3bc8492dbdf439e0b7efa831439c0c7b341f7

PR #113, opened 2015-08-22
Disable tests that fail on big endian architectures due to crc mismatches. (#112)

sebastic:
CRS tests are now only run on little endian architectures like `test/t/geom/test_wkb.cpp`


joto:
CRC should be independent of byte ordering now. If it isn't than thats a bug that needs fixing.

PR #114, opened 2015-08-23
do not cache the version string

DerDakon:
All components are unconditionally set before, so there is no need to put the
version string in the cache. Doing so would pin the version string in case
someone updates the source without deleting the build directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,9 +30,7 @@ set(LIBOSMIUM_VERSION_MINOR 3)
 set(LIBOSMIUM_VERSION_PATCH 0)
 
 set(LIBOSMIUM_VERSION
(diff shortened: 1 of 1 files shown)


joto:
Hm. I did this on purpose at some point (d4b48ebb7f688a2667f35bbb6d2ac43eab97d12c). Not sure why any more. I guess I should write better commit messages :-). I guess I wanted to be able to change the version dynamically. Or at least get it displayed. Does this make sense?

DerDakon:
You mean displayed in the cache? Well, this sort-of makes sense, but this has the potential of breaking things when someone updates his checkout but will still get a build with the old version string. Since the packaging (which uses the component variables) and the string may show different versions I fear this could create more trouble than it is worth.

PR #119, opened 2015-08-25
only enable C++ language

DerDakon:
CMake enables both C and C++ by default, but since libosmium only uses C++ code
there is no need to check for the C compiler.


joto:
Fails on Windows. See appveyor test.

PR #121, opened 2015-09-08
FindOsmium: prevent errors in list(REMOVE_DUPLICATES) when no components are requested

DerDakon:


joto:
This change will always set `OSMIUM_INCLUDE_DIRS` even if `OSMIUM_FOUND` is not set which I vaguely remember isn't how such a module is supposed to work? Maybe it doesn't matter?

DerDakon:
If OSMIUM_FOUND is false the content of the other variables is more or less undefined. Other modules (including those shipped with CMake itself) also return data in those variables if *_FOUND is false, e.g. because headers were found, but the version does not match the requested one.

PR #125, opened 2015-09-24
Added protozero installation;

ilidar:
Just updated CMakeLists.txt to install protozero (at least for OSX);


joto:
I am not sure how to handle this. Protozero is an external library and might already be installed from somewhere else. Maybe this should be a separate target? Or a CMake option a la INSTALL_PROTOZERO_TOO?

ilidar:
@joto yes, you're right, implicit install could cause problems;

Issue #124, opened 2015-09-23
Improve Travis CI setup mainly wrt. modern compiler coverage


daniel-j-h:
This is a master ticket for tracking and discussing improving the Travis CI integration.

- ~~[ ] fail fast, `set -e` on current scripts~~
- [x] integrate different compilers, release, debug builds

I'm currently playing with improving the OSRM Travis CI integration (as we are not even using the new container infrastructure right now). I'm hoping to re-use some of this integration work here.

Playground is here: https://travis-ci.org/daniel-j-h/cpp-travis-setup
and: https://github.com/daniel-j-h/cpp-travis-setup/blob/master/.travis.yml

daniel-j-h:
First results are here:
https://travis-ci.org/daniel-j-h/libosmium

This is running from my fork at:
https://github.com/daniel-j-h/libosmium

in which I basically just modified the `.travis.yml` config.

daniel-j-h:
Next goals:

- [ ] [ASan / TSan / MSan / UBSan](https://github.com/Project-OSRM/osrm-backend/issues/1601) builds

daniel-j-h:
Unfortunately the Sanitizers from the deb packages are broken right now, bug here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=779785

daniel-j-h:
@joto would love to get your thoughts on this: if you want me to change something, if this is a good start, or what you need before considering merging it.

joto:
@daniel-j-h: interesting work. Shame we have the problems on travis with the sanitizers, but I guess this will be fixed some day. Lets keep this open and revisit once the debian packages are fixed on travis.

daniel-j-h:
This landed in https://github.com/osmcode/libosmium/commit/4ad6e4340e8ef8cf0f50e1783e6d6c16d5cfa737 closing here.

PR #126, opened 2015-09-25
Travis CI setup improvements: more compilers, caching, etc.

daniel-j-h:
As tracked in https://github.com/osmcode/libosmium/issues/124.

I really really wanted the sanitizers to work, but unfortunately was stumbling across various issues.
This is definitely something we have to integrate in the future!


joto:
The pull request contains several unrelated changes, includes whitespace changes and removes a, to my mind useful, comment at the beginning of the travis file for no apparent reason. Can you split this up into several pieces? I am happy to find better ways to format code etc., but it is much easier for me to see whats going on when functionality changes and formatting changes are separate.

daniel-j-h:
Sorry for that. This mainly comes from me building the Travis config completely from scratch [here](https://github.com/daniel-j-h/cpp-travis-setup) and then incorporating the small libosmium related changes e.g. for packages and dependencies.

I did it this way, since I structured the file in a different way than the old libosmium Travis file.

I also use some YAML tricks to make life easier: `&` and `*` is used for anchoring, to generate aliases as to not repeat ourselves. `|` is used for multiline commands.

Let me guide you through the new Travis config.
Read along at: https://github.com/daniel-j-h/libosmium/blob/travis_improvements/.travis.yml

With `sudo=false` we activate the new container based infrastructure (enables caching, faster startups, 2 cores, 4gb ram, etc.), but we lose the sudo privileges. That's okay though.

The next lines are for caching. Right now we only use `ccache`, but we could cache arbitrary directories (hint: data test files in the future?) to speed up the builds. The ccache settings are self-explanatory, the CPP2 config is for ccache to work with Clang.

We then build the test matrix. Unfortunately, this still has to be done manually, since there were some open Travis issues. We use YAML anchoring, building some aliases to not repeat ourselves.

The matrix consists basically of the following combinations:
- os
- compiler
- build type

That is, we build for both, Linux and OSX. On Linux we then build for various recent GCC and Clang versions. On OSX we only use Clang, but build for several Xcode releases.

The sources specify the apt sources to fetch the packages from, since Travis is based on Ubuntu 12.04 with outdated packages. The `ubuntu-toolchain-r-test` provides GCC/stdlibc++, the llvm sources the LLVM/Clang counterparts.

The reason we have to set COMPILER and later use it, instead of setting CXX is that Travis by default overwrites the CXX env var with the compiler we have to specify as the compiler key (without the version).

At the bottom: in the install step, we grab a recent CMake (again, Travis has an outdated one, Ubuntu 12.04), and on osx install the dependencies via brew (as there are no apt sources there).

We then run cmake, ctest and hope everything works out :P

The result is here: https://travis-ci.org/daniel-j-h/libosmium

daniel-j-h:
Closing here, development is going on in https://github.com/osmcode/libosmium/pull/132 after talking to @joto.

PR #157, opened 2016-06-02
Replace std::atof to avoid locale-dependence

tinloaf:
Hi,

with std::atof, the question of what is interpreted as a decimal separator depends on the system's locale. However, the XML files always use a dot. Thus, if trying to parse such an XML file for example on a German system, all coordinates are being cut down to their integer parts.

Replacing atof with the iostream-parsing is not only "more C++", but also provides the possibility to change the locale of that specific stream to "C".

### DIFF
--- a/include/osmium/io/detail/xml_input_format.hpp
+++ b/include/osmium/io/detail/xml_input_format.hpp
@@ -126,6 +126,14 @@ namespace osmium {
 
         namespace detail {
 
+            double atof_helper(const char *str) {
+              double val = 0.0f;
+              std::istringstream istr(str);
+              istr.imbue(std::locale("C"));
+              istr >> val;
+              return val;
+            }
+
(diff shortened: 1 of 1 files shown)


joto:
Unfortunately iostream is also horribly inefficient, so that's not a good option here either. `atof` uses the process locale, but we can't change that due to multithreading. Looks like we are stuck unless we do it ourselves...

tinloaf:
Would you accept a hacky but fast solution? In this case, I'd figure out what the locale's decimal separator is (only need to do that once...) and replace every "." in the string by that...

joto:
I have solved the problem in a different way. I have written my own functions for converting those location coordinates. This is now in master. It also solves the reverse problem of writing out floating point values to XML, OPL or debug files. Thanks for pointing out this problem and thanks for your patch even if I didn't use it in the end.

Issue #165, opened 2016-09-30
Failed parsing of coordinates missing leading 0


karenzshea:
I'm seeing an issue on libosmium 2.9.0 with parsing coordinates that are provided missing a leading digit, as in `.05`, instead of say `0.05`. The error is thrown:

```
terminate called after throwing an instance of 'osmium::invalid_location'
  what():  wrong format for coordinate: '.05'
During startup program terminated with signal SIGABRT, Aborted.
```

Fixing the coordinate it's parsing to `0.05` fixes it.

cc @joto

PR #166, opened 2016-10-04
Support coordinates missing leading 0 digit

karenzshea:
Closes https://github.com/osmcode/libosmium/issues/165.

Previous logic for parsing coordinates assumes that the leading character will be a digit and not the optional decimal. I added logic to keep track of how many integers have been parsed.

I also updated Catch up a patch version; let me know if that should go in a separate PR.


joto:
Thanks for the fix, but I found one that looks simpler. And yes, updating catch should go into its own commit. Will do that, too, now.

PR #177, opened 2016-11-18
Markdown fixes

pnorman:


joto:
Do I see this correctly? The patch adds an invisible space at the end of line? Maybe a better fix would be to add a `*` at the beginning of the line making this a list?

pnorman:
> Do I see this correctly? The patch adds an invisible space at the end of line?

Two spaces at the end of a line is how you do a newline in Markdown. I didn't want to change doc, just fix the markdown, but I can change it to a list.

joto:
Yes, please do. Invisible spaces looks like a really bad idea to me.

PR #179, opened 2016-12-13
Add match_key<std::regex> tag

oxidase:
Added a tag `match_key<std::regex>` for regexp key filters `osmium::tags::Filter<std::regex>`


joto:
Thanks @oxidase. Accepted despite the wrong formatting. Next time please use the same formatting as in the rest of the code.

oxidase:
Oh, it is the OSRM formatting style, i have to be careful with project styles.

PR #186, opened 2016-12-22
Add docs for releasing + npm publishing

springmeyer:
Because libosmium is header-only, it is very easy to include in various packaging/build systems. A common usecase at Mapbox is writing node c++ modules that depend on header-only libraries. 

Like https://github.com/mapbox/protozero/pull/59 this adds a few small things to libosmium to make it easy to:

  - publish libosmium to npm
  - allow any node addon to dynamically download and compile against the code.

This makes sense for seamless inclusion of libosmium in node modules. As a proof of concept of this usage I've written a node module that pulls in libosmium as a dependency [here](https://github.com/springmeyer/node-osmium-count/blob/2db99f458350dadab55db234c59bd797f199cce0/package.json#L17) and then dynamically builds against it [here](https://github.com/springmeyer/node-osmium-count/blob/2db99f458350dadab55db234c59bd797f199cce0/binding.gyp#L16) (possible due to the `include_dirs.js` functionality).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,3 +10,7 @@ Some rules for contributing to this project:
   you have added. Please read the [notes for developers](NOTES_FOR_DEVELOPERS.md)
   beforehand which contains some coding guidelines.
 
+
+### Releasing
+
+See https://github.com/osmcode/osmium-suite/blob/master/make_releases.md
--- a/include_dirs.js
+++ b/include_dirs.js
@@ -1,0 +1,2 @@
+var path = require('path');
+console.log(path.join(path.relative('.', __dirname),'include'));
--- a/package.json
+++ b/package.json
@@ -1,0 +1,13 @@
+{
+    "name": "libosmium",
+    "version": "2.10.3",
+    "description": "Fast and flexible C++ library for working with OpenStreetMap data",
+    "main": "include_dirs.js",
+    "repository"   :  {
+      "type" : "git",
+      "url"  : "git://github.com/osmcode/libosmium.git"
+    },
+    "files": [
+        "include/osmium"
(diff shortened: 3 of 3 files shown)


joto:
Release instructions for libosmium are [here](https://github.com/osmcode/osmium-suite/blob/master/make_releases.md). Because we are usually releasing together with other software of the suite, it makes sense to have this in one place.

The Makefile in libosmium repos is out of data and will be removed. Don't add anything there.

I don't understand the need for this, isn't this done by node-osmium already?

springmeyer:
@joto - thanks for the pointers. I will revise the PR now that I know where the release instructions are - sorry I missed those.

> The Makefile in libosmium repos is out of data and will be removed. Don't add anything there.

Okay.

> I don't understand the need for this, isn't this done by node-osmium already?

This will allow node-osmium (and other node c++ modules that I envision writing using libosmium directly) to more easily integrate the libosmium headers.  So, once this is done I will be able to remove this line (https://github.com/osmcode/node-osmium/blob/e6b915cf56879fd0bb6a38d54a6dd53ef17cecb9/.travis.yml#L101) and instead list `libosmium` as a `dependency`, which is cleaner, clearer, and less brittle to maintain.

springmeyer:
Rebased and ready for review again.

joto:
I don't think the source repository is the right place for having this specialized packaging config. All other packaging (Debian, Fedora, Homebrew) is done outside the repository in their own place somewhere. Although in this case I might be the one releasing the new npm packages myself, generally packaging is a different job than development, done by different people with different priorities and constraints. So I see this as something outside the core libosmium.

springmeyer:
Agreed, closing.

PR #190, opened 2017-01-26
Gentoo ebuild

andrew-aladev:
Added a gentoo ebuild.


joto:
Package management stuff should be done somewhere else, not in the source repository. There are many package managers and we can't accommodate them all. If it makes sense for Gentoo, I am willing to consider having a special repository under the osmcode user for this, if there is no other place this can be put.

andrew-aladev:
Ok, please create this repository.

joto:
https://github.com/osmcode/osmium-gentoo

PR #199, opened 2017-03-03
Don't look up the node location if it is already known.

Nakaner:
If the input file has extra `lat` and `lon` attributes in the `<nd>` tags, this lookup is not necessary.

This fixes the following bug.

The fixed bug
-------------------

Assume the following OSM XML file (called `testfile.osm`):
```xml
<?xml version='1.0' encoding='UTF-8'?>
<osm version='0.6' upload='true' generator='vim'>
  <way id='1' visible='true'>
    <nd ref='1' lat='50.59752093205' lon='6.24685043376'/>
    <nd ref='2' lat='50.59825880818' lon='6.24848414852' />
    <nd ref='3' lat='50.59844327541' lon='6.25113108061' />
    <tag k='highway' v='secondary' />
  </way>
</osm>
```

and following Osmium program:
```c++
#include <iostream>
#include <osmium/geom/wkt.hpp>
#include <osmium/index/map/sparse_mmap_array.hpp>
#include <osmium/handler/node_locations_for_ways.hpp>
#include <osmium/io/reader.hpp>
#include <osmium/io/xml_input.hpp>
#include <osmium/osm/way.hpp>
#include <osmium/visitor.hpp>
#include <osmium/handler.hpp>

using index_type = osmium::index::map::SparseMmapArray<osmium::unsigned_object_id_type, osmium::Location>;
using location_handler_type = osmium::handler::NodeLocationsForWays<index_type>;

class MyHandler : public osmium::handler::Handler {
    osmium::geom::WKTFactory<> m_factory;

public:
    void way(const osmium::Way& way) {
        std::cout << m_factory.create_linestring(way) << "\n";
    }
};

int main(int, char**) {
    MyHandler handler;

    // This variant is necessary if you want to read OSM files which do not contain node locations on ways.
    // It does not work with files which contain node locations on ways but not the nodes themselves.
    osmium::io::Reader reader1("testfile.osm");
    try {
        index_type index;
        location_handler_type location_handler{index};
        osmium::apply(reader1, location_handler, handler);
        std::cout << "option 1 worked\n";
    } catch (osmium::not_found&) {
        std::cout << "option 1 failed\n";
    }
    reader1.close();

    // This variant is necessary if you want to read OSM files which contain node locations on ways but not the nodes themselves.
    // It does not work with files without node locations on ways because of the lacking location handler.
    osmium::io::Reader reader2("testfile.osm");
    try {
        osmium::apply(reader2, handler);
        std::cout << "option 2 worked\n";
    } catch (osmium::not_found&) {
        std::cout << "option 2 failed\n";
    }
    reader2.close();
}
```

The expected output is:
```
LINESTRING(6.2468504 50.5975209,6.2484841 50.5982588,6.2511311 50.5984433)
option 1 worked
LINESTRING(6.2468504 50.5975209,6.2484841 50.5982588,6.2511311 50.5984433)
option 2 worked
```

But I get the following using the current master branch (commit 3e0f6fc):
```
option 1 failed
LINESTRING(6.2468504 50.5975209,6.2484841 50.5982588,6.2511311 50.5984433)
option 2 worked
```

Build this example with:
```
g++ -std=c++11  -Wall -Wextra  -O3 -g -c -I /home/michael/git/libosmium/include -o waylocationstest.o waylocationstest.cpp
g++ -std=c++11  -Wall -Wextra  -O3 -g -o waylocationstest.o  waylocationstest.cpp -lexpat -lpthread -lz
```

### DIFF
--- a/include/osmium/handler/node_locations_for_ways.hpp
+++ b/include/osmium/handler/node_locations_for_ways.hpp
@@ -155,6 +155,11 @@ namespace osmium {
                 }
                 bool error = false;
                 for (auto& node_ref : way.nodes()) {
+                    if (node_ref.location()) {
+                        // Node location is already known because the input file contained it as extra
+                        // attributes in the <nd> tag.
+                        continue;
+                    }
                     try {
                         node_ref.set_location(get_node_location(node_ref.ref()));
                         if (!node_ref.location()) {


joto:
Hm. Not sure about this. If you are using the `NodeLocationsForWays` handler you get what you asked for. Why use the handler at all if you don't need it? I can only see that this would make sense if some way nodes have the lat/lon on it and some haven't. But does this actually happen in a real-world case?

Thinking about how your change could break things: Is there a usage scenario where you specifically want to ignore what's in the file and use the data from the node location store?

Nakaner:
If you want to write a program which read OSM files, does something and needs the location of the nodes, you usually have a `NodeLocationsForWays` handler because the node references do not have latitude and longitude information. But if you want to read an OSM file with the same program but the file has latitude and longitude information on the ways and no nodes (because they are not necessary), an `osmium::not_found` exception is thrown. Throwing an catching those exception costs time. Checking if an OSM file contains way locations on all node references requires an additional reading of the OSM file.

This patch allows to use the same code for both OSM files without way locations and OSM files with way locations but no nodes inside.

@joto  wrote:
> Thinking about how your change could break things: Is there a usage scenario where you specifically want to ignore what's in the file and use the data from the node location store?

I had the same concern in mind while writing this patch. I think that a file where the location information on the way differs from the nodes should be considered as broken.

Is the following a solution you would accept? If yes, I will add it to this PR. By calling the constructor of `NodeLocationsForWays` a user can decide how to handle locations on ways_

* ignore way locatons
* use them before looking the location up in the index
* looking it up in the index and using the way locations as a fallback

What should be the default? I think that (2) should be the default. Shall I declare the old constructor as deprecated?

joto:
If you don't need the `NodeLocationHandler` it adds a huge memory overhead. The whole point of having the node locations on the ways is that this overhead is avoided. So using the NodeLocationHandler in your code when you don't need it, doesn't make sense. This isn't so bad if you are using one of the `Sparse` indexes, but if you are using any of the `Dense` indexes, your index will even have the same size, only there is not much in it.

The real solution here is to have a branch in your code that doesn't use the `NodeLocationHandler` at all. This can be triggered by a configuration option or command line option. When using PBF files, the information whether this file contains locations on way nodes is stored in the header option `LocationsOnWays`. You can check whether this flag is set. This flag is not available for XML files, because, unlike with PBF files, there is no clear place where to put it.

PR #204, opened 2017-04-14
fix CMake build if Ruby is not available

alex85k:
If there is no Ruby in the path, cmake just shows an error on finding Gems (tested on Windows). Better to skip data tests as usual.


joto:
Wrong variable was checked. 3cdb15eb0455c707e85c86e132f4232b07e6e112 should fix it.

alex85k:
Oh, yes, it was ``find_program``, not ``find_package(Ruby)``. Thank you for fixing.

PR #213, opened 2017-07-12
Fix deadlock during Pool shutdown at (after) shared library unloading

mloskot:
## Use case

libosmium used in a shared library

## Problem:

At exit of a client using such shared library,
Pool is destroyed during static storage clean-up.
Typically, that is after shared library has been unloaded.
Then, `Pool::shutdown_all_workers` goes into deadlock waiting on threads to exit (a known tricky thread synchronization issue).

## Solution:

Let libosmium user *work around* the deadlock calling (public) `Pool::shutdown_all_workers`, at convenient time before shared library is unloaded, forcing Pool to skip shutting down threads (which have already been shut) during Pool singleton destruction.

------

Notice, I suggest it as a workaround.
The advantage is that the interface does not change to anything more complex, an average user should not notice any difference.

There may be a better/proper solution though, certainly.

### DIFF
--- a/include/osmium/thread/pool.hpp
+++ b/include/osmium/thread/pool.hpp
@@ -116,6 +116,7 @@ namespace osmium {
             std::vector<std::thread> m_threads;
             thread_joiner m_joiner;
             int m_num_threads;
+            bool m_all_workers_down;
 
             void worker_thread() {
                 osmium::thread::set_thread_name("_osmium_worker");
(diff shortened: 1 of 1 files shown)


mloskot:
Thanks @joto for the further explanation.

Obviously, I have missed the fact shutdown does not join the threads.

This makes the workaround incorrect.

mloskot:
Closed as failed attempt.
Better solution is proposed in #214

PR #214, opened 2017-07-12
Enable use of user-provided thread::Pool instance

mloskot:
Enable use of user-provided `thread::Pool` instance

`Pool` instance can be configured with user-specified number of threads and passed to `Reader` and `Writer` as optional argument of constructors.
If no `Pool` is specified, default instance is used and configured with number of threads based on actual number of cores on the system.

Advantages:
* Allows to use `Pool` configured with user-preferred number of threads.
* If libosmium used in a shared library, this helps to prevent potential deadlock during Pool shutdown at (after) shared library unloading (see #213 for details).

------
* Compiled with GCC 7.1.0 and passed all tests on Ubuntu Linux (complete test suite)
* Compiled with VS2017 and run using custom sample reader/writer (XML, PBF) app

### DIFF
--- a/include/osmium/io/detail/output_format.hpp
+++ b/include/osmium/io/detail/output_format.hpp
@@ -46,6 +46,7 @@ DEALINGS IN THE SOFTWARE.
 #include <osmium/io/file.hpp>
 #include <osmium/io/file_format.hpp>
 #include <osmium/memory/buffer.hpp>
+#include <osmium/thread/pool.hpp>
 
 namespace osmium {
 
@@ -107,6 +108,7 @@ namespace osmium {
 
             protected:
 
+                osmium::thread::Pool& m_pool;
                 future_string_queue_type& m_output_queue;
 
                 /**
@@ -119,7 +121,8 @@ namespace osmium {
 
             public:
 
-                explicit OutputFormat(future_string_queue_type& output_queue) :
+                OutputFormat(osmium::thread::Pool& pool, future_string_queue_type& output_queue) :
+                    m_pool(pool),
                     m_output_queue(output_queue) {
                 }
 
@@ -152,7 +155,7 @@ namespace osmium {
 
             public:
 
-                using create_output_type = std::function<osmium::io::detail::OutputFormat*(const osmium::io::File&, future_string_queue_type&)>;
+                using create_output_type = std::function<osmium::io::detail::OutputFormat*(osmium::thread::Pool&, const osmium::io::File&, future_string_queue_type&)>;
 
             private:
 
@@ -178,10 +181,10 @@ namespace osmium {
                     return true;
                 }
 
-                std::unique_ptr<osmium::io::detail::OutputFormat> create_output(const osmium::io::File& file, future_string_queue_type& output_queue) {
+                std::unique_ptr<osmium::io::detail::OutputFormat> create_output(osmium::thread::Pool& pool, const osmium::io::File& file, future_string_queue_type& output_queue) {
                     auto it = m_callbacks.find(file.format());
                     if (it != m_callbacks.end()) {
-                        return std::unique_ptr<osmium::io::detail::OutputFormat>((it->second)(file, output_queue));
+                        return std::unique_ptr<osmium::io::detail::OutputFormat>((it->second)(pool, file, output_queue));
                     }
 
                     throw unsupported_file_format_error(
--- a/include/osmium/io/detail/pbf_input_format.hpp
+++ b/include/osmium/io/detail/pbf_input_format.hpp
@@ -183,7 +183,7 @@ namespace osmium {
                         PBFDataBlobDecoder data_blob_parser{std::move(input_buffer), read_types(), read_metadata()};
 
                         if (osmium::config::use_pool_threads_for_pbf_parsing()) {
-                            send_to_output_queue(osmium::thread::Pool::instance().submit(std::move(data_blob_parser)));
+                            send_to_output_queue(get_pool().submit(std::move(data_blob_parser)));
                         } else {
                             send_to_output_queue(data_blob_parser());
                         }
--- a/include/osmium/io/writer.hpp
+++ b/include/osmium/io/writer.hpp
@@ -53,6 +53,7 @@ DEALINGS IN THE SOFTWARE.
 #include <osmium/io/header.hpp>
 #include <osmium/io/writer_options.hpp>
 #include <osmium/memory/buffer.hpp>
+#include <osmium/thread/pool.hpp>
 #include <osmium/thread/util.hpp>
 #include <osmium/util/config.hpp>
 #include <osmium/version.hpp>
@@ -169,7 +170,12 @@ namespace osmium {
                 osmium::io::Header header;
                 overwrite allow_overwrite = overwrite::no;
                 fsync sync = fsync::no;
+                osmium::thread::Pool* pool = nullptr;
             };
+
+            static void set_option(options_type& options, osmium::thread::Pool& pool) {
+                options.pool = &pool;
+            }
 
             static void set_option(options_type& options, const osmium::io::Header& header) {
                 options.header = header;
@@ -223,7 +229,7 @@ namespace osmium {
             explicit Writer(const osmium::io::File& file, TArgs&&... args) :
                 m_file(file.check()),
                 m_output_queue(detail::get_output_queue_size(), "raw_output"),
-                m_output(osmium::io::detail::OutputFormatFactory::instance().create_output(m_file, m_output_queue)),
+                m_output(nullptr),
                 m_buffer(),
                 m_buffer_size(default_buffer_size),
                 m_write_future(),
@@ -235,6 +241,12 @@ namespace osmium {
                 (void)std::initializer_list<int>{
                     (set_option(options, args), 0)...
                 };
+
+                if (!options.pool) {
+                    options.pool = &thread::Pool::default_instance();
+                }
+
+                m_output = osmium::io::detail::OutputFormatFactory::instance().create_output(*options.pool, m_file, m_output_queue);
 
                 if (options.header.get("generator") == "") {
                     options.header.set("generator", "libosmium/" LIBOSMIUM_VERSION_STRING);
--- a/include/osmium/thread/pool.hpp
+++ b/include/osmium/thread/pool.hpp
@@ -130,6 +130,11 @@ namespace osmium {
                 }
             }
 
+        public:
+
+            static constexpr int default_num_threads = 0;
+            static constexpr int default_queue_size = 0;
+
             /**
              * Create thread pool with the given number of threads. If
              * num_threads is 0, the number of threads is read from
@@ -141,9 +146,12 @@ namespace osmium {
              * given number, ie it will leave a number of cores unused.
              *
              * In all cases the minimum number of threads in the pool is 1.
+             *
+             * If max_queue_size is 0, the queue size is read from
+             * the environment variable OSMIUM_MAX_WORK_QUEUE_SIZE.
              */
-            explicit Pool(int num_threads, size_t max_queue_size) :
-                m_work_queue(max_queue_size, "work"),
+            explicit Pool(int num_threads = default_num_threads, size_t max_queue_size = default_queue_size) :
+                m_work_queue(max_queue_size > 0 ? max_queue_size : detail::get_work_queue_size(), "work"),
                 m_threads(),
                 m_joiner(m_threads),
                 m_num_threads(detail::get_pool_size(num_threads, osmium::config::get_pool_threads(), std::thread::hardware_concurrency())) {
@@ -158,12 +166,8 @@ namespace osmium {
                 }
             }
 
-        public:
-
-            static constexpr int default_num_threads = 0;
-
-            static Pool& instance() {
-                static Pool pool(default_num_threads, detail::get_work_queue_size());
+            static Pool& default_instance() {
+                static Pool pool{};
                 return pool;
             }
 
@@ -176,6 +180,10 @@ namespace osmium {
 
             ~Pool() {
                 shutdown_all_workers();
+            }
+
+            int num_threads() const noexcept {
+                return m_num_threads;
             }
 
             size_t queue_size() const {
--- a/test/t/io/test_writer.cpp
+++ b/test/t/io/test_writer.cpp
@@ -111,3 +111,40 @@ TEST_CASE("Writer") {
 
 }
 
+TEST_CASE("Writer with user-provided pool") {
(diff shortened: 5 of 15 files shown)


test/t/thread/test_pool.cpp
```diff
@@ -70,3 +70,30 @@ TEST_CASE("thread") {
...
+    SECTION("can send job to thread pool") {
+        auto future = pool.submit(test_job_with_result {});
+
+        REQUIRE(future.get() == 42);
+    }
+
+    SECTION("can throw from job in thread pool") {
+        auto future = pool.submit(test_job_throw {});
+
+        REQUIRE_THROWS_AS(future.get(), std::runtime_error);
```
joto: Should be `std::runtime_error&`.

joto:
I think to be consistent, we should also do the same for `Reader`. But first get it right for the `Writer`.

include/osmium/io/writer.hpp
```diff
@@ -236,9 +243,12 @@ namespace osmium {
                     (set_option(options, args), 0)...
                 };
 
-                if (options.header.get("generator") == "") {
-                    options.header.set("generator", "libosmium/" LIBOSMIUM_VERSION_STRING);
-                }
```
mloskot: I was preparing the patch using 2.12.2, then diff-ported to master, loosing some post-2.12.2 changes, apparently. I'll restore it.

include/osmium/io/writer.hpp
```diff
@@ -236,9 +243,12 @@ namespace osmium {
...
+                if (options.pool)
+                    m_pool = options.pool;
+                else
+                    m_pool = &thread::Pool::default_instance();
+
+                m_output = osmium::io::detail::OutputFormatFactory::instance().create_output(*m_pool, m_file, m_output_queue);
```
mloskot: Your concern seems similar to the situation discussed here:
https://akrzemi1.wordpress.com/2012/11/14/not-using-stdthread/#comment-659

Isn't `Writer::m_thread` considered to actually be *"launching of the thread is the last instruction in the constructor body"* (if we shuffle `ensure_cleanup` a bit) ?

include/osmium/thread/pool.hpp
```diff
@@ -162,7 +162,11 @@ namespace osmium {
 
             static constexpr int default_num_threads = 0;
 
-            static Pool& instance() {
+            static std::unique_ptr<Pool> create_instance(int num_threads = default_num_threads) {
+                return std::unique_ptr<Pool>(new Pool(num_threads, detail::get_work_queue_size()));
+            }
```
mloskot: > Do we need this function?

No, unless `Pool` ctor must be non-public, what I assumed.

I also aimed for some interface symmetry, more explicit and self-describing that one is supposed to use either `default_instance`  or `create_instance`, not mixing them.

If you prefer, I can remove it and make `Pool` ctor public. Let me know.

include/osmium/thread/pool.hpp
```diff
@@ -178,6 +182,10 @@ namespace osmium {
                 shutdown_all_workers();
             }
 
+            int num_threads() const {
+                return m_num_threads;
```
mloskot: BTW, `queue_size`  and `queue_empty` are not specified with `noexcept`. Is it intentional or the refactoring is in progress?

include/osmium/io/detail/pbf_input_format.hpp
```diff
@@ -183,7 +183,7 @@ namespace osmium {
                         PBFDataBlobDecoder data_blob_parser{std::move(input_buffer), read_types(), read_metadata()};
 
                         if (osmium::config::use_pool_threads_for_pbf_parsing()) {
-                            send_to_output_queue(osmium::thread::Pool::instance().submit(std::move(data_blob_parser)));
+                            send_to_output_queue(osmium::thread::Pool::default_instance().submit(std::move(data_blob_parser)));
```
mloskot: Hmm, I'm trying to look closer and I can't see anything broken, comparing to the file in the current master.

pbf_input_format hpp

include/osmium/io/detail/output_format.hpp
```diff
@@ -129,6 +132,7 @@ namespace osmium {
                 OutputFormat& operator=(const OutputFormat&) = delete;
                 OutputFormat& operator=(OutputFormat&&) = delete;
 
+
```
mloskot: Unintentional. I'll fix it.

include/osmium/thread/pool.hpp
joto: There is really no reason why the constructor can't be public. So please change that.

test/t/io/test_writer.cpp
```diff
@@ -111,3 +111,56 @@ TEST_CASE("Writer") {
...
+    osmium::memory::Buffer buffer = reader.read();
+    REQUIRE(buffer);
+    REQUIRE(buffer.committed() > 0);
+    const auto num = std::distance(buffer.select<osmium::OSMObject>().cbegin(), buffer.select<osmium::OSMObject>().cend());
+    REQUIRE(num > 0);
+    REQUIRE(buffer.select<osmium::OSMObject>().cbegin()->id() == 1);
+
+    std::string filename;
+
+    SECTION("Successfull writes") {
```
mloskot: A perfect copy-n-paste mistake :)

Fixed the first one in #215

test/t/io/test_writer.cpp
```diff
@@ -111,3 +111,56 @@ TEST_CASE("Writer") {
...
+
+        SECTION("Writer buffer") {
```
mloskot: Copying the original test was a quick hack.

No problem, I will refactor once we agree what to do about `create_instance` vs ctor.

test/t/io/test_writer.cpp
```diff
@@ -111,3 +111,56 @@ TEST_CASE("Writer") {
...
+        }
+
+        osmium::io::Reader reader_check{filename};
+        osmium::memory::Buffer buffer_check = reader_check.read();
+        REQUIRE(buffer_check);
+        REQUIRE(buffer_check.committed() > 0);
+        REQUIRE(buffer_check.select<osmium::OSMObject>().size() == num);
+        REQUIRE(buffer_check.select<osmium::OSMObject>().cbegin()->id() == 1);
+    }
+}
```
mloskot: Interesting. I suspect if this is not caused by git, due to [config.autocrlf=true](https://help.github.com/articles/dealing-with-line-endings/#global-settings-for-line-endings).

I don't use any auto-trimming in editor.

test/t/thread/test_pool.cpp
```diff
@@ -70,3 +70,30 @@ TEST_CASE("thread") {
...
+    SECTION("can get access to thread pool") {
+        REQUIRE(pool.queue_empty());
+    }
+    
+    SECTION("can access user-provided number of threads") {
+        REQUIRE(pool.num_threads() == 7);
+    }
+
+    SECTION("can send job to thread pool") {
+        auto future = pool.submit(test_job_with_result {});
```
mloskot: The formatting is not mine, [it was there](https://github.com/osmcode/libosmium/blob/6a17a8ddd58c6d574259ac839fb550540e017eee/test/t/thread/test_pool.cpp#L60).

test/t/thread/test_pool.cpp
mloskot: Same here, I copied [what was there](https://github.com/osmcode/libosmium/blob/6a17a8ddd58c6d574259ac839fb550540e017eee/test/t/thread/test_pool.cpp#L68), didn't I?

mloskot:
> I think to be consistent, we should also do the same for Reader.

Yes.

> But first get it right for the Writer.

I like this approach.

test/t/thread/test_pool.cpp
joto: Sorry, not your fault. My code was wrong. I'll fix my code. :-)

mloskot: No problem, cheers :-)

test/t/io/test_writer.cpp
mloskot: Should be fixed in the latest update 1d051f909b5d4a0aa5c76191e27300ff998398cc

include/osmium/thread/pool.hpp
mloskot: I've removed `create_instance`  in 1d051f909b5d4a0aa5c76191e27300ff998398cc

test/t/thread/test_pool.cpp
mloskot: I also fixed mine.

test/t/thread/test_pool.cpp
mloskot: I also fixed mine.

test/t/io/test_writer.cpp
mloskot: Here is the current update and I have no idea why GitHub is not picking up the change (added new line) and not marking this review comment as outdated.

https://github.com/mloskot/libosmium/blob/a113611a206908823ca8931bfbd8e0bdf346372a/test/t/io/test_writer.cpp#L167

mloskot:
@joto I think I've completed the first major round of review. If you could have a look at the updates and see what else needs improving.

include/osmium/thread/pool.hpp
```diff
@@ -162,7 +162,11 @@ namespace osmium {
 
             static constexpr int default_num_threads = 0;
 
-            static Pool& instance() {
+            explicit Pool(int num_threads = default_num_threads) :
```
joto: Having a forwarding constructor here doesn't buy us anything. Just move the existing constructor down here so it becomes public and make the second parameter default to `detail::get_work_queue_size()`.

include/osmium/thread/pool.hpp
```diff
@@ -162,7 +162,11 @@ namespace osmium {
...
+                Pool(num_threads, detail::get_work_queue_size()) {
+            }
+
+            static Pool& default_instance() {
                 static Pool pool(default_num_threads, detail::get_work_queue_size());
```
joto: This then becomes just `static Pool pool{}`.

test/t/io/test_writer.cpp
```diff
@@ -111,3 +111,43 @@ TEST_CASE("Writer") {
...
+TEST_CASE("Writer with user-provided Pool") {
+    osmium::io::Header header;
+    header.set("generator", "test_writer.cpp");
+
+    osmium::io::Reader reader{with_data_dir("t/io/data.osm")};
+    osmium::memory::Buffer buffer = reader.read();
+    REQUIRE(buffer);
+    REQUIRE(buffer.committed() > 0);
+
+    SECTION("Successful writes") {
```
joto: Extra SECTION over everything doesn't help here, probably just slows down compilation.

test/t/thread/test_pool.cpp
```diff
@@ -48,9 +48,32 @@ TEST_CASE("number of threads in pool") {
 
 }
 
+TEST_CASE("number of threads in user-provided pool") {
+
+     SECTION("if zero requested, threads configured") {
```
joto: These should all be just TEST_CASEs without any SECTIONs. There is no common code here.

test/t/io/test_writer.cpp
mloskot: I decided to keep it to state clearly there are no `Pool` tests that exercise failing writer. IOW, `Pool` must be sound.
I can it remove though.

test/t/thread/test_pool.cpp
mloskot: I will change it to TEST_CASEs

include/osmium/thread/pool.hpp
mloskot: It buys simpler interface. unless it is desired to let user to specify the work queue too, what I have just learned.

(note, in this PR my concern is *number of threads*, other aspects of pool may deserve separate PR, who knows)

mloskot: Besides, `detail::*` used in public interface smells.

Anyhow, I'll change.

test/t/io/test_writer.cpp
joto: These tests are just about the `Pool` so that's okay if there is nothing failing in here, we don't need to say this explicitly.

include/osmium/thread/pool.hpp
joto: Good point! How about this: make the default value `0`. Check inside the constructor if it is and then use the `detail::get_work_queue_size()`. The the documentation should say: "Set to 0 to get max queue size from OSMIUM_MAX_WORK_QUEUE_SIZE env variable". The the detail is hidden behind a good default value with the added benefit that we'll never have an actual queue size of 0 which would make no sense.

test/t/io/test_writer.cpp
mloskot: Agreed

mloskot:
Second round of review requests addressed.

mloskot:
https://travis-ci.org/osmcode/libosmium/jobs/253639146

```
 57/115 Test  #57: io_test_writer_with_mock_encoder .........***Exception: SegFault  0.04 sec
```

Hmm, only one Xcode builds is failing. I don't have access to OSX to investigate if this is actually related to this PR.

joto:
The test failure on OSX is probably something else. The master branch also shows the same problem.

I have looked at the `Writer` constructor in more detail now and I think it is mostly fine the way you wrote it. Just initialize `m_output(nullptr)` so we have the complete list of members. And we don't need `m_pool`, just do a  `if (!options.pool) { options.pool = &thread::Pool::default_instance() }` and use `options.pool` then when creating the output.

joto:
Now that the writer side is clear, lets look at the reader side, too. This should be even easier, because of the [struct parser_arguments](https://github.com/osmcode/libosmium/blob/master/include/osmium/io/detail/input_format.hpp#L58-L64) which can easily be extended with a `Pool` member which will then be available in all the right places. Can you add this to your pull request?

mloskot:
@joto https://github.com/osmcode/libosmium/pull/214#issuecomment-315464056
> Just initialize m_output(nullptr) so we have the complete list of members (...)

Done.

mloskot:
@joto https://github.com/osmcode/libosmium/pull/214#issuecomment-315465792

First stab at `Pool`-enabled `Reader` is in.

Since only `PBFParser`-based reader actually uses the pool, if there was a .pbf test data added, the `Reader` test case with user-provided `Pool` could be switched to use .pbf.

joto:
Thanks @mloskot for the great work. I have merged the pull request now. If needed, we can do further refinement later.

mloskot:
@joto I'm glad we could help. Thank you for your assistance.

Issue #224, opened 2017-08-21
_WIN32 vs. _MSC_VER


joto:
The libosmium code contains some places with `#ifdef _WIN32` and others with `#ifdef _MSC_VER`. According to https://github.com/cxong/tinydir/issues/18, which I just stumbled upon, this is probably false.
Shall we change all `_MSC_VER` to `_WIN32` or are there some cases where `_MSC_VER` is actually correct?

/cc @BergWerkGIS @alex85k

BergWerkGIS:
Well, there is no wrong or right.
Both are correct - each for their purpose.

* `_WIN32` is for generally checking if the app is built for/on Windows.
* `_MSC_VER` is specifically targeted towards the Microsoft compiler (aka Visual Studio or C++ Build Tools) and checking the version thereof as each version might have different ~~bugs~~ aehm **features** :smirk:

```
MSVC++ 4.x  _MSC_VER == 1000
MSVC++ 5.0  _MSC_VER == 1100
MSVC++ 6.0  _MSC_VER == 1200
MSVC++ 7.0  _MSC_VER == 1300
MSVC++ 7.1  _MSC_VER == 1310 (Visual Studio 2003)
MSVC++ 8.0  _MSC_VER == 1400 (Visual Studio 2005)
MSVC++ 9.0  _MSC_VER == 1500 (Visual Studio 2008)
MSVC++ 10.0 _MSC_VER == 1600 (Visual Studio 2010)
MSVC++ 11.0 _MSC_VER == 1700 (Visual Studio 2012)
MSVC++ 12.0 _MSC_VER == 1800 (Visual Studio 2013)
MSVC++ 14.0 _MSC_VER == 1900 (Visual Studio 2015)
MSVC++ 14.1 _MSC_VER == 1910 (Visual Studio 2017)
```

joto:
Sure, but how would I know? All the Windows code I have been writing came either from somebody else or I have been guessing from docs and snippets here or there. Can we try compiling with clang or gcc on Windows and see what works and what doesn't?

alex85k:
As far as I see, there are no _MSC_VER specific version checks now (we assume Visual Studio 2015 or later to have base C++ 11).
Many of the checks are including `unistd.h` file with POSIX api, not available on Windows but having some equivalents in `io.h`  or C++ 11. 
I can check with MSYS2 (latest GCC on Windows).

joto:
[...]

PR #226, opened 2017-09-11
Make libosmium compatible with MSYS2 (gcc 7.2.0 on Windows)

alex85k:
As mentioned in https://github.com/osmcode/libosmium/issues/224 , both _WIN32 and _MSC_VER are used in the code to support Windows build. On Visual Studio the compiler is different and needs more workarounds (no POSIX library function). Basic _WIN32 checks are used for Unx-specific or non-C++-standard functions, like file IO and memory mapping.

However, some checks were incorrectly placed and GCC compiler on Windows (working as part of MSYS2 environment) failed to build libosmium tests and examples. Here is my attemt to fix it (Microsoft compatibility is not broken, just checked, all tests pass).

### DIFF
--- a/include/osmium/osm/timestamp.hpp
+++ b/include/osmium/osm/timestamp.hpp
@@ -230,11 +230,11 @@ namespace osmium {
 #ifndef NDEBUG
                 auto result =
 #endif
-#ifndef _MSC_VER
-                              gmtime_r(&sse, &tm);
+#ifndef _WIN32
+                gmtime_r(&sse, &tm);
                 assert(result != nullptr);
 #else
-                              gmtime_s(&tm, &sse);
+                gmtime_s(&tm, &sse);
                 assert(result == 0);
 #endif
 
--- a/include/osmium/util/file.hpp
+++ b/include/osmium/util/file.hpp
@@ -174,7 +174,7 @@ namespace osmium {
          * @throws std::system_error If ftruncate(2) call failed
          */
         inline void resize_file(int fd, std::size_t new_size) {
-#ifdef _WIN32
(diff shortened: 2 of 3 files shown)


joto:
The `gmtime_s` commit is wrong. It should use `gmtime_r` on Unix. I think you didn't see the `n` in `ifndef`.

alex85k:
Oops, still mixing ifdef and ifndef) Sorry, fixing.

alex85k:
Hope that this will fix it. MSYS2 is almost fully compatible with Linux, it even has cppcheck, doxygen and clang-tidy avalable. However, there are some nasty bugs in libraries (such as common PROJ.4 hardcoded path error and broken mktemp()) that result in some failing tests.

joto:
Thanks @alex85k . I am merging your changes now. If there is a way to bring this into appveyor it would help keeping things working. But I see this as low priority, because I don't know of any users who actually need this.

Concerning the failing tests you mentioned: If you can see some easy workarounds for us so we don't hit those library problems, we could think about them. But if they are not something a real user would likely hit its probably not worth it.

mloskot:
@joto 
> If there is a way to bring this into appveyor it would help keeping things working. 

If you mean add build job based on MinGW/MSYS, the [environment is there](https://www.appveyor.com/docs/build-environment/#mingw-msys-cygwin) and here is sample config based on CMake and MinGW 23 adn 64 bit: https://github.com/lexicalunit/nanodbc/blob/master/appveyor.yml

As libosmium user on Windows, I thank @alex85k for the clean-up.

alex85k:
Will try to make appveyor - msys2 variant now.

PR #227, opened 2017-09-15
Add Msys2 building appveor

alex85k:
Following https://github.com/osmcode/libosmium/issues/224  and https://github.com/osmcode/libosmium/pull/226

Here is the build on Windows with latest MinGW compiler provided by MSYS2 project. The dependencies are also provied and downloaded as binary pacakages (may be updating, versions are not fixed).

I have also found a source of minor problem with not-finding json ruby gem - gem executable was called gem.cmd and was not found with FindGem.cmake.
With current version of MSYS2 packages all the tests should pass.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,12 +6,14 @@
 
 environment:
   matrix:
+  - config: MSYS2
+    autocrlf: true
   - config: Dev
+    autocrlf: true
+  - config: RelWithDebInfo
     autocrlf: true
   - config: Dev
     autocrlf: false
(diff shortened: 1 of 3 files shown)


joto:
I have merged this, thank you!

There are some warnings in the appveyor log about `winsock2.h` and `windows.h`. Do we need to do something there?

alex85k:
The best thing to do about that warnings is to eliminate winsock headers from pbf reader. The only need for them is `ntohl` function here. Will make PR with this simplification (now that we have protozero with similar functionality)

PR #228, opened 2017-09-16
Fix testdata-multipolygon test on Windows (executable path differs)

alex85k:
Noticed that for MSVC builds the testsdata-multipolygon test failed (previously it was not run because of gem finding bug). The reason was /Release/ folder in test executable path (tests are run with `ctest -VV -C Release` but our manual code did not use that rare option). So the solution proposed is to pass EXECUTABLE parameter to run-testdata-multipolygon.cmake . Should make appveyor even more happy :)


joto:
I have merged this. But the test was successful in MSYS2 build anyway and is still disabled for MSVC build due to missing ruby json module.

joto:
I would it expect to not work, because that test also needs spatialite support in GDAL which is not available. That's why the `testdata-overview` is disabled in the appveyor build, it needs spatialite support in GDAL.

PR #229, opened 2017-09-16
No winsock dependency

alex85k:
Many warnings and previous build problems were related to winsock2 library we used. But currently the only place where it is needed is byte order conversion https://github.com/osmcode/libosmium/blob/381cf92723a1fab5b1fc74fcb8f0ad89d73b7274/include/osmium/io/detail/pbf_output_format.hpp#L188
Now that we have protozero library, this can easily be avoided, at least on Windows (little-endian only).


alex85k:
(accidentally included commit from previous PR)

mloskot:
This is pretty cool (given previous issues like #198)

PR #230, opened 2017-09-21
Proposed Geometry Factory Improvements: Polygons, GeoJSON, and RapidGeoJSONDocument Interface

worace:
Hi there -

I'm hoping to contribute some improvements to some of the GeometryFactory implementations in libosmium.

This is my first time contributing to libosmium and I've also not had a ton of experience writing C++ so I have tried to follow along with the conventions used in the library as best as I could, but I'd be happy to receive any advice or improvements you would like to recommend.

Here's a brief rundown of the issues I was running into with the library and the changes I'm proposing to deal with them:

## Proposed Changes

There are a few different changes included in this so I'll try to summarize each one:

### 1. Adding "polygon" implementation to existing WKT and GeoJSON Factories

I was trying to use these classes in another project and noticed there did not seem to be an implementation of the various Polygon handler methods for the WKT and GeoJSON writers. I'm afraid I don't have an exact example of the error this produces available any more but roughly the WKT or GeoJSON writer would error out because there is no `polygon_start` implementation included. I added the necessary methods for this along with test cases to cover them.

It's worth noting that the one I added so far is a fairly naive Polygon implementation in that it does not expect any of the geometries to include interior rings. However based on my understanding of how OSM represents these geometries I think this is OK since those would always be handled as Areas and thus get serialized as MultiPolygons anyway.

### 2. Fixing the existing RapidGeoJSON implementation and adding tests

Based on what I was observing trying to use this class it seems that the existing version doesn't work because:
* **a)** We were adding an unnecessary `geometry` key to the value (which is only needed if producing a geojson Feature rather than just a geometry.
* **b)** We weren't inserting a root object in the rapidjson document. Rapidjson requires starting a root object before you can start adding key/value members. I hopefully fixed this and added some tests for that case as well.

One note here is that I also discovered that the existing RapidGeoJSON writer implementation is only good for writing a single object; after that rapidjson will complain about multiple root values in the document, unless the user explicitly resets their stream and writer.

I think it would be possible to handle this more automatically in the library but it would probably require changing some of the interface in a breaking way so for now I left it as it is and documented it with a test.

### 3. Adding a new `RapidGeoJSONDocumentFactory` Geometry Factory

This is the change I'm personally most excited about so I hope you will agree that it seems like a reasonable addition for the library.

Basically I wanted a version of the GeometryFactory that would return `rapidjson::Document` objects so that I could then do additional manipulations on those structures before I serialize them to strings.

With the existing implementation I would have needed to get a JSON string back from the factory, re-parse it, and then do whatever I needed.

So this new version which I'm hoping to add (along with tests) covers the case of getting an un-serialized rapidjson value out.


worace:
One thing I realize will need to be sorted out is how to add the rapidjson dep to the appveyor build (seeing as it just failed).

I attempted to add it for the travis build so I'll see after the next run if that worked, but I was pretty lost looking at all the appveyor config, so might need to solicit some instruction on how to configure that.

mloskot:
> One thing I realize will need to be sorted out is how to add the rapidjson dep to the appveyor build (seeing as it just failed).

```
find_package(Git REQUIRED)
ExternalProject_Add(
  rapidjson
  PREFIX ${CMAKE_BINARY_DIR}/rapidjson
  GIT_REPOSITORY https://github.com/Tencent/rapidjson.git
  TIMEOUT 10
  UPDATE_COMMAND ${GIT_EXECUTABLE} pull
  CONFIGURE_COMMAND ""
  BUILD_COMMAND ""
  INSTALL_COMMAND ""
  LOG_DOWNLOAD ON)
ExternalProject_Get_Property(catch source_dir)
set(RAPIDJSON_INCLUDE_DIR ${source_dir}/rapidjson/include/rapidjson CACHE INTERNAL "Path to include folder for RapidJSON")
```

or, might be simpler, install from VCPKG package (see https://www.appveyor.com/docs/lang/cpp/)

``
vcpkg install rapidjson
```

then integrate in one of two ways, see https://github.com/Microsoft/vcpkg/blob/master/docs/users/integration.md

joto:
Thanks @worace for your efforts. First, this should be split up into multiple pull requests to make it easier to discuss and handle each part separately. You have already nicely identified three parts which do different things, so just put those in separate pull requests. But read the following first...

To point 1: Polygons created from closed ways only are rarely useful just by themselves because you are missing the data from multipolygon relations which gives you an incomplete picture of the OSM data. That's why I haven't built out the polygon callbacks in all the geometry handlers. Some have it as a sort of legacy because at some point before Osmium had multipolygon support it was all we had. Do you have an actual use case where you need those polygons?

To point 2: I have used the RapidGeoJSON implementation successfully, so it is not broken (but maybe not right for your use case). It is supposed to generate the "geometry part" of a GeoJSON structure that can be then used as part of a larger GeoJSON feature that you are supposed to create yourself outside libosmium. You can see in https://github.com/osmcode/osmium-tool/blob/master/src/export/export_format_json.cpp how it is meant to be used.

To point 3: This seems to me a rather specific use case, especially because keeping the whole JSON tree in memory would only work for rather small files. So at the moment I don't think this is something for the library. But it is a totally valid thing to do and the class can easily live outside the library and be just as useful. One reason the geometry builders are designed the way they are is that everybody can plug in their own. I could see it as something for [osmium-contrib](https://github.com/osmcode/osmium-contrib).

worace:
@joto thanks for that feedback, I'm going to close this and open new PRs. I'll respond to your comments there so we can go through it 1 piece at a time.

PR #232, opened 2017-09-22
Add polygon implementation for WKT and GeoJSON Geometry Factories

worace:
I was trying to use these classes in another project and noticed there did not seem to be an implementation of the various Polygon handler methods for the WKT and GeoJSON writers. I'm afraid I don't have an exact example of the error this produces available any more but roughly the WKT or GeoJSON writer would error out because there is no polygon_start implementation included. I added the necessary methods for this along with test cases to cover them.

It's worth noting that the one I added so far is a fairly naive Polygon implementation in that it does not expect any of the geometries to include interior rings. However based on my understanding of how OSM represents these geometries I think this is OK since those would always be handled as Areas and thus get serialized as MultiPolygons anyway.

### DIFF
--- a/include/osmium/geom/geojson.hpp
+++ b/include/osmium/geom/geojson.hpp
@@ -98,6 +98,28 @@ namespace osmium {
                     return str;
                 }
 
+                /* Polygon */
+                void polygon_start() {
+                    m_str = "{\"type\":\"Polygon\",\"coordinates\":[[";
+                }
+
+                void polygon_add_location(const osmium::geom::Coordinates& xy) {
+                    xy.append_to_string(m_str, '[', ',', ']', m_precision);
+                    m_str += ',';
+                }
+
+                polygon_type polygon_finish(size_t /* num_points */) {
+                    assert(!m_str.empty());
+                    std::string str;
+
+                    using std::swap;
+                    swap(str, m_str);
+
+                    str.back() = ']';
+                    str += "]}";
+                    return str;
+                }
+
                 /* MultiPolygon */
 
                 void multipolygon_start() {
--- a/include/osmium/geom/wkt.hpp
+++ b/include/osmium/geom/wkt.hpp
@@ -110,6 +110,29 @@ namespace osmium {
                     return str;
                 }
 
+                /* Polygon */
+                void polygon_start() {
+                    m_str = m_srid_prefix;
+                    m_str += "POLYGON((";
+                }
+
+                void polygon_add_location(const osmium::geom::Coordinates& xy) {
+                    xy.append_to_string(m_str, ' ', m_precision);
+                    m_str += ',';
+                }
+
+                polygon_type polygon_finish(size_t /* num_points */) {
+                    assert(!m_str.empty());
+                    std::string str;
+
+                    using std::swap;
+                    swap(str, m_str);
+
+                    str.back() = ')';
+                    str += ")";
+                    return str;
+                }
+
                 /* MultiPolygon */
 
                 void multipolygon_start() {
--- a/test/t/geom/test_geojson.cpp
+++ b/test/t/geom/test_geojson.cpp
@@ -80,6 +80,32 @@ TEST_CASE("GeoJSON linestring geometry") {
 
 }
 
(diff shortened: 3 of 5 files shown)


worace:
@joto here's the new PR for just the polygon implementation that I mentioned in my other PR.

In response to your comment:

> To point 1: Polygons created from closed ways only are rarely useful just by themselves because you are missing the data from multipolygon relations which gives you an incomplete picture of the OSM data. That's why I haven't built out the polygon callbacks in all the geometry handlers. Some have it as a sort of legacy because at some point before Osmium had multipolygon support it was all we had. Do you have an actual use case where you need those polygons?

I'd love to understand this better since I may well be confused about how OSM models these, but I had the impression that while there are some closed ways that comprise MultiPolygon relations, the vast majority of closed polygons (e.g. buildings, etc) are modeled in OSM just as a way where the first and last nodes are the same.

To give just a single example here is a building that is comprised of a single closed way but not (as far as I can tell) a member of any Relations: http://www.openstreetmap.org/way/224889715

Here's an example of that polygon showing up in Mapbox: http://jsfiddle.net/8u1sc4pf/5/ which I'm assuming is coming out of OSM, so presumably they (and probably other users) are consuming these single-way-non-relation polygons somehow.

So on the one hand it seems to me to make sense that the library would include a means of serializing these ways for consumption in other GIS tools, but on the other hand they're obviously getting it done somehow at the moment so perhaps there is an implementation I missed somewhere.

Thanks again for your attention on this!

joto:
When you are using the area support of osmium, you'll get both polygons from closed ways as well as (multi)polygons from relations as what Osmium calls Areas. Those will then be converted to multipolygons by the GeometryFactory. This way they are handled both in the same way transparently, any user of Osmium doesn't have to know where an Area came from, they are all the same. Areas created from closed ways will simply just have a single outer ring. The "polygon" functions in the GeometryFactory *only* handle the closed ways, but not the multipolygon relations, so you'll only get *some* areas, which is almost never what you want.

worace:
@joto so when you say
> When you are using the area support of osmium
Do you mean this class? https://github.com/osmcode/libosmium/blob/master/include/osmium/area/multipolygon_manager.hpp

I was working with a modified version of that in the tool I have been trying to make but my experience was that I didn't seem to get the normal-closed-way-polygons unless I added this: https://github.com/worace/osm_denorm/blob/master/cpp/osm_denorm.cxx#L214-L219

However it's very likely I'm doing something wrong in how I'm attempting to use the multipolygon manager so maybe you will recognize what issue I'm having there.

joto:
Please don't use this PR to ask general questions unrelated to the PR. There are plenty of examples how to use the area support in Osmium in the examples directory or in https://github.com/osmcode/osm-gis-export/blob/master/src/osmium_toogr2.cpp for instance. Have a look there and compare it to your code. You definitely don't need the `way_not_in_any_relation` function.

worace:
You are right, sorry for the digression and thanks for the info you provided! I've looked through some more examples in the osmium-tool export commands and I understand now why the Polygon implementation has not been needed thus far.

As far as OSM Areas go it seems like in an ideal world the areas that have only a single outer ring would be output as Polygons rather than MultiPolygons, but the current approach obviously works since most systems don't take any objection to a multipolygon that could also be represented as a polygon.

The other argument in my mind for implementing these is that currently there is an external interface for `create_polygon` which fails for the geojson and wkt implementations since they don't have the required internal methods to do the work.

Perhaps you think it will be more misleading to have these in there at all (since it would make people think, as I did, they need to use the `create_polygon` interface when in fact everything is being done with multipolygons. I'll defer to you on that count.

So in the end this feels like a small nice-to-have in terms of completeness but if you think there are reasons to keep it out I understand.

joto:
I'll have to think about this some more. You are right in that it doesn't make much sense to have this implemented for some geometry factories, but not for others. Either I'll accept your PR and have them for all, or I need to mark these functions as *deprecated* in the other factories that have them.

The question of whether to generate Polygon geometries or MultiPolygon geometries for geometries with a single outer ring is a different one. The reasoning there is simple: It is easier for most users of the data to handle just one consistent type.

PR #233, opened 2017-09-22
Support standalone geometries from RapidGeoJSON factory

worace:
@joto here's a separate PR for part 2 of what I was mentioning in https://github.com/osmcode/libosmium/pull/230

I understand better how you are using this class after looking at the example you shared (https://github.com/osmcode/osmium-tool/blob/master/src/export/export_format_json.cpp) -- thanks for sharing that!

Personally as a user I found the interface a bit confusing so I wonder if you'd be open to still considering whether to automatically include that geometry key or not.

My thinking on it is that the existing interface only supports the usage you described (embedding geometry into a pre-started GeoJSON Feature). However it seems like with this change we would be able to support both use-cases: writing to an empty stream without requiring the caller to do any setup ahead of time **as well as** allowing the caller to embed in a GeoJSON Feature if that is what they want.

I have included a test here that shows an example of this use case: https://github.com/osmcode/libosmium/compare/master...worace:rapidgeojson-standalone-geometries?expand=1#diff-2251f2d972fad0336ed206aa99193021R268

However you definitely have more experience working with these interfaces so I'd love to know your thoughts and would defer to you if there are reasons you think the current interface is better.

Even if you think it's best to leave the interface as is I do feel like it would be nice to include some documentation or tests demonstrating how the class is supposed to be used. I'd be happy to update the test cases I added here to demonstrate that calling convention if you think that is the way to go.


joto:
There are several issues here. First, we can't just do any change we might want, because we have to be backwards compatible. You change isn't backwards compatible, so we can't do it this way. But there are probably ways around this, introducing some optional parameter or so. The next question is what this change buys us? What do you need this for that's different then inserting it in a GeoJSON file? Can you elaborate on your use case?

And yes, more tests and documentation are always needed. I am aware that Osmium is a complex library that isn't well documented. Any help is appreciated.

worace:
@joto sure, I realize this is a breaking change so I would not have expected it to go in until whatever next osmium release that was acceptable for. But if you see that as too large of a break I totally understand.

To step back a bit my original use-case which got me started investigating all of this stuff was just to make an export tool for denormalizing and converting OSM geometries into something like GeoJSON or WKT.

I'm actually learning a lot from this blog post that I just found through your site: https://blog.jochentopf.com/2017-08-30-osmium-releases.html, as it seems like you may have added something to the osmium tool that actually does what I've been trying to achieve.

So that is great!

It looks like this feature of the tool is fairly new, which makes sense as I am pretty sure I looked around for something similar a couple months ago but did not find anything. We had been using this other tool which is based on an alternate implementation in Go: github.com/pelias/pbf2json, but it does not account for the Area assembly so would miss some data for Relations.

I'm experimenting with the new osmium export now some more and it looks pretty great.

One thing I am noticing is many closed ways seem to still come out as LineString, for example this one: http://www.openstreetmap.org/way/4336012

Came out as:

```json
{"type":"Feature","geometry":{"type":"LineString","coordinates":[[-122.0550718,37.4140975],[-122.0550772,37.4140246],[-122.0550219,37.4140205],[-122.0550289,37.4139746],[-122.0550217,37.413911],[-122.0538658,37.411656],[-122.0537842,37.4115954],[-122.053812,37.4115579],[-122.0537141,37.4115075],[-122.0535981,37.4114716],[-122.0534692,37.4114582],[-122.0533268,37.4114683],[-122.0531895,37.4115091],[-122.0530677,37.4115681],[-122.0529763,37.4116346],[-122.0529006,37.4117257],[-122.0528492,37.4118647],[-122.0529061,37.4118758],[-122.0529002,37.4119247],[-122.0529066,37.4119847],[-122.0540481,37.4142228],[-122.0540987,37.4142698],[-122.0541482,37.4142997],[-122.0541191,37.4143384],[-122.0542618,37.4144042],[-122.0544174,37.4144298],[-122.0545837,37.4144341],[-122.0547392,37.4143957],[-122.0548787,37.4143318],[-122.0549914,37.4142338],[-122.0550718,37.4140975]]},"properties":{"name":"Hangar One","image":"http://commons.wikimedia.org/wiki/File:Last_Look_at_Hangar_One.jpg","aeroway":"hangar","building":"yes","wikipedia":"en:Hangar One (Mountain View, California)","start_date":"1933"}}
```

When I exported with the tool. Is that what you would expect to see there? Or is it possible to get those as actual geojson polygons?

joto:
If you have questions etc. not related to this pull request, please either create new issues in the appropriate repositories or ask on the [dev mailinglist](https://lists.openstreetmap.org/listinfo/dev). Issues and pull requests are not a general discussion forum.

worace:
Sure thing. Going to close this for now since it seems like changing the existing behavior would be problematic. If you think there would be value in the additional tests I'd be happy to rework those to match the current implementation rather than the one I described in this PR.

The obvious downside there is it would require adding rapidjson as a dependency to the test environment, so perhaps you can let me know if you have strong feelings there one way or another.

joto:
@worace Having tests is always good, so I'd be happy to include any tests you create for the current behaviour. I'd have to think about the extra dependency. This has come up with other libraries, too, and I haven't found the right solution there yet. I'd probably include it as an optional dependency, so if CMake finds rapidjson it will include the tests and if not, then they are simply not compiled.

PR #243, opened 2018-02-06
Implement new metadata options for PBF, OPL and debug output format

Nakaner:


include/osmium/io/detail/debug_output_format.hpp
```diff
@@ -220,25 +221,37 @@ namespace osmium {
...
                     output_int(object.id());
                     *m_out += '\n';
-                    if (m_options.add_metadata) {
+                    if (m_options.add_metadata.version()) {
                         write_fieldname("version");
                         *m_out += "  ";
                         output_int(object.version());
+                    }
+                    if (m_options.add_metadata.any()) {
                         if (object.visible()) {
```
joto: If `version` is off, this will probably look wrong with a leading space before `visible` or `deleted`. Have to check the ouput.

Nakaner: You are right.

With version:
```
header
  multiple object versions: no
  bounding boxes: 
  options: 
    generator = osmium/1.7.1
    version = 0.6

=============================================

node 1
  version:   4 visible
  user:      2 "foo"
  lon/lat:   10,50

```

Without version:

```
header
  multiple object versions: no
  bounding boxes: 
  options: 
    generator = osmium/1.7.1
    version = 0.6

=============================================

node 1
 visible
  user:      2 "foo"
  lon/lat:   10,50
```

Shall I move visibility on its own line? I think that we don't have to preserve backward compatibility here because this format is intended to be read by a human and not processed by a program (like OPL).

include/osmium/io/detail/pbf_output_format.hpp
```diff
@@ -487,7 +515,11 @@ namespace osmium {
                     m_primitive_block(m_options) {
                     m_options.use_dense_nodes = file.is_not_false("pbf_dense_nodes");
                     m_options.use_compression = file.get("pbf_compression") != "none" && file.is_not_false("pbf_compression");
-                    m_options.add_metadata = file.is_not_false("pbf_add_metadata") && file.is_not_false("add_metadata");
+                    m_options.add_metadata = metadata_options{file.get("add_metadata")};
```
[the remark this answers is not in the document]
Nakaner: I can change it but the old and the new behaviour would differ then. Up to now both `pbf_add_metadata` and `add_metadata` have to be true (both default to true) in order to write metadata. If I first check `pbf_add_metadata` and `add_metadata` afterwards, `add_metadata=true` (default) can overwrite `pbf_add_metadata=false`. That's why I would use

```cpp
if (!file.is_not_false("pbf_add_metadata") || file.get("pbf_add_metadata") == "none") {
    m_options.add_metadata = metadata_options{"none"};
} else {
    m_options.add_metadata = metadata_options{file.get("add_metadata")};
}
```

include/osmium/io/detail/debug_output_format.hpp
joto: Yeah. Lets move it behind the first line:
```
    node 1 visible
      version:   4
      user:      2 "foo"
```

include/osmium/io/detail/pbf_output_format.hpp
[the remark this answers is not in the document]
joto: Okay, this is getting way to complicated for my little head. :-) Lets do this differently: If `pbf_add_metadata` is set in any way, we throw an `invalid_argument` exception: "The 'pbf_add_metadata' option is deprecated. Please use 'add_metadata' instead." And then we just interpret the `add_metadata`. No sense thinking about the interaction between those two parameters when one is deprecated anyway. I doubt anybody is still using it anyway.

Nakaner:
All issues should be fixed now.

PR #248, opened 2018-03-09
Extend osmium::io::detail::metadata_options, add unit tests and move it out of detail namespace

Nakaner:
Changes:

* Move from namespace `osmium::io::detail::metadata_options` to `osmium::io::metadata_options`. I think that's ok because there has not been any stable release since it has been added to Osmium. The class [will be used](https://github.com/osmcode/osmium-tool/pull/91) by Osmium Tool, i.e. an "external" application and it is no implementation detail any more.
* Make the internal enum of `osmium::io::metadata_options` a typed enum (char).
* Add an output of the string representation (needed by Osmium Tool).
* Improve test coverage.


test/t/io/test_metadata.cpp
```diff
@@ -87,6 +91,87 @@ TEST_CASE("Metadata options: timestamp,uid,user") {
...
+
+TEST_CASE("Metdata options: constructor using OSMObject") {
+    osmium::memory::Buffer buffer{10 * 1000};
+    using namespace osmium::builder::attr;
+
+    SECTION("only version") {
+        const osmium::OSMObject& obj = buffer.get<osmium::OSMObject>(osmium::builder::add_node(buffer, _id(1),
+                _version(2)));
+        osmium::io::metadata_options options{obj};
+        REQUIRE(!options.user());
```
joto: Please use `REQUIRE_FALSE(...)` instead of `REQUIRE(!...)`.

include/osmium/io/metadata_options.hpp
```diff
@@ -0,0 +1,204 @@
...
+        template <typename TChar, typename TTraits>
+        inline bool write_option(std::basic_ostream<TChar, TTraits>& out, bool already_written, const char* name) {
+            if (already_written) {
+                out << '+';
+            }
+            out << name;
+            return true;
+        }
+
+        template <typename TChar, typename TTraits>
```
joto: I suggest writing a `metadata_options_to_string` function that returns a std::string. There you can assemble the string by always appending the metadata option name and the `+` sign, last thing then is to `s.resize(s.size() - 1)` to remove the last `+`. Makes the code much simpler and can be useful if you need a string that you don't want to output to a stream. The ostream operator is a one-liner then.

The tests you have below are much simpler to write, too.

Nakaner: You mean that [method](https://github.com/osmcode/libosmium/pull/248/commits/31e34f10c52f08a53469e06e98867982bf31560a#diff-195c657e6eb2642212801e6799d5dd78R159) which I removed in my last commit?

joto: Yep. That's it. :-)

Also one tip there: If you move the `std::string result` to the beginning and set it even in the `all` and `none` case, the same `std::string` is always used as the return value which should allow return value optimization to work which would remove a copy.

include/osmium/io/metadata_options.hpp
```diff
@@ -0,0 +1,204 @@
...
+#include <stdexcept>
+#include <string>
+#include <osmium/osm/object.hpp>
+
+namespace osmium {
+
+    namespace io {
+
+        /**
+         * Stores information on which metadata attributes should be
```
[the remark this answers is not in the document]
Nakaner: The only alternative which came into my mind: "Stores which metadata attributes are available in an OSM file or should be written to an OSM file."

include/osmium/io/metadata_options.hpp
```diff
@@ -92,27 +92,7 @@ namespace osmium {
...
-                // to distinguish them from objects with a reduced number of metadata fields.
-                if (object.uid() > 0) {
-                    opts |= options::md_uid;
-                }
-                if (object.user() && strlen(object.user()) > 0) {
-                    opts |= options::md_user;
-                }
-                m_options = static_cast<options>(opts);
-            }
+            friend metadata_options metadata_options_from_object(const osmium::OSMObject& object);
```
joto: No friends please. The point of making this an extra function is to decuple internals from functions that don't need it. You'll need to add setter functions to metadata_options for this.

include/osmium/io/metadata_options.hpp
```diff
@@ -196,6 +176,33 @@ namespace osmium {
             return out << options.to_string();
         }
 
+        /**
+         * Create an instance of metadata_options from an instance of osmium::OSMObject
```
joto: This could better describe what this function is for.

include/osmium/io/metadata_options.hpp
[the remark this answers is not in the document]
joto: This isn't strictly about a file any more, you are using it also to describe which metadata an OSMObject has. Maybe something like "Describes which metadata fields are available in an OSMObject. This is most often used to describe which metadata fields are available in the objects of an OSM file."

Nakaner:
I addressed all your suggestions (I hope I didn't oversee any).

test/t/io/test_metadata.cpp
```diff
@@ -1,9 +1,13 @@
 #include "catch.hpp"
 
-#include <osmium/io/detail/metadata_options.hpp>
+#include <sstream>
```
joto: Not needed any more.

test/t/io/test_metadata.cpp
```diff
@@ -1,9 +1,13 @@
...
+#include <osmium/builder/attr.hpp>
+#include <osmium/io/metadata_options.hpp>
+#include <osmium/memory/buffer.hpp>
+#include <osmium/osm/object.hpp>
 
 TEST_CASE("Metadata options: default") {
-    osmium::io::detail::metadata_options m;
+    osmium::io::metadata_options m;
```
joto: Should use `const` where possible.

include/osmium/io/metadata_options.hpp
```diff
@@ -0,0 +1,239 @@
...
+            bool none() const noexcept {
+                return m_options == 0;
+            }
+
+            bool version() const noexcept {
+                return (m_options & options::md_version) != 0;
+            }
+
+            void set_version(bool flag) noexcept {
+                if (flag) {
```
joto: Can we do these without the conditional? I just checked this on https://godbolt.org/ and only the newest GCCs can optimize away the conditional.

include/osmium/io/metadata_options.hpp
```diff
@@ -0,0 +1,239 @@
...
+#include <stdexcept>
+#include <string>
+#include <osmium/osm/object.hpp>
+
+namespace osmium {
+
+    namespace io {
+
+        /**
+         * Describes which metadata fields are available in an OSMObject. This is most often used to
```
joto: Now that I am seeing this being not so much about IO any more, I think we should move this out of the `io` namespace directly into the `osmium` namespace. And the file would move (again) to the `include/osmium/osm/` directory I think.

include/osmium/io/metadata_options.hpp
```diff
@@ -0,0 +1,239 @@
...
+SHALL THE COPYRIGHT HOLDERS OR ANYONE DISTRIBUTING THE SOFTWARE BE LIABLE
+FOR ANY DAMAGES OR OTHER LIABILITY, WHETHER IN CONTRACT, TORT OR OTHERWISE,
+ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+
+*/
+
+#include <osmium/util/string.hpp>
+
+#include <stdexcept>
```
joto: We also need `#include <ostream>`

test/t/io/test_metadata.cpp
```diff
@@ -87,6 +91,78 @@ TEST_CASE("Metadata options: timestamp,uid,user") {
...
+        REQUIRE(options.timestamp());
+        REQUIRE(options.changeset());
+        REQUIRE_FALSE(options.user());
+        REQUIRE_FALSE(options.uid());
+    }
+}
+
+TEST_CASE("Metdata options: string representation should be valid") {
+
+    SECTION("version+changeset") {
```
joto: Can you make these SECTIONs into their own TEST_CASEs? There is no common code in the TEST_CASE, so no need for the hierarchy here.

PR #254, opened 2018-05-04
Add option use_default_metadata_values

Nakaner:
Many applications including Libosmium < 2.14.0 and Osmosis require either all or no metadata fields in PBF files. They fail reading files which contain only version and timestamp fields. This pull request adds a new option `use_default_metadata_values=true` to write default values (`user="" uid=0 changeset=0`) to make these applications happy.

### DIFF
--- a/include/osmium/osm/timestamp.hpp
+++ b/include/osmium/osm/timestamp.hpp
@@ -238,11 +238,13 @@ namespace osmium {
          * Return the timestamp as string in ISO date/time
          * ("yyyy-mm-ddThh:mm:ssZ") format. If the timestamp is invalid, an
          * empty string will be returned.
+         *
+         * @param output_zero output the string even if it is 1970-01-01T00:00:00Z
          */
-        std::string to_iso() const {
+        std::string to_iso(const bool output_zero = false) const {
             std::string s;
 
-            if (m_timestamp != 0) {
+            if (m_timestamp != 0 || output_zero) {
                 struct tm tm;
                 time_t sse = seconds_since_epoch();
 #ifndef NDEBUG
--- a/include/osmium/io/detail/pbf_output_format.hpp
+++ b/include/osmium/io/detail/pbf_output_format.hpp
@@ -87,6 +87,9 @@ namespace osmium {
                 /// Which metadata of objects should be added?
                 osmium::metadata_options add_metadata;
 
+                /// Write zeros to those metadata fields which should not be written?
+                bool default_metadata_values = false;
+
                 /// Should nodes be encoded in DenseNodes?
                 bool use_dense_nodes = true;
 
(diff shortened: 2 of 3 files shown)


joto:
I have thought about this some more and I don't think this is something that should be added to libosmium. Essentially what you are doing here is changing those metadata values to some value (which happens to be the default). You can do that in an application, you don't need library support for this. Not writing out metadata values needs library support, but writing them out doesn't.

There might be an issue here with the zero timestamp, so maybe we need a second function which writes the `1970-01-01T00:00:00Z` timestamp, to stay compatible in the XML case. I'd prefer a second function to the "magic bool" approach, because I can't think of a reason why we might want to make this configurable at runtime.

An easy workaround for you: Write out the OSM file without metadata and then read that and write it out again without metadata.

Nakaner:
I wrote this feature to produce OSM files which are usable by a large number of programs who did not or do not treat all fields in `DenseInfo` messages as optional (as specified). It is an error that these applications don't do what they should do but forcing all users [1] to upgrade their software isn't feasible either.

> An easy workaround for you: Write out the OSM file without metadata and then read that and write it out again without metadata.

That results in either writing the file without user, uid and changeset field, reading it again and writing it with all metadata fields or in writing a pipe construct like this one:

```
osmium cat --output-format pbf,add_metadata=version+timestamp with_all.osm.pbf | osmium cat --output-format pbf,add_metadata=all -o nulls.osm.pbf
```

Where should it go if it shouldn't be implemented in the library? In Osmium-Tool? I think that that approach is worse. I would have to use a builder for every object just to copy the object from the input buffer because read objects are `const` and `set_changeset()` isn't `const` for good reasons. Not to mention the code bloat of it (even if writing objects were moved into its own class).

include/osmium/osm/timestamp.hpp
```diff
@@ -238,11 +238,13 @@ namespace osmium {
...
          * empty string will be returned.
+         *
+         * @param output_zero output the string even if it is 1970-01-01T00:00:00Z
          */
-        std::string to_iso() const {
+        std::string to_iso(const bool output_zero = false) const {
             std::string s;
 
-            if (m_timestamp != 0) {
+            if (m_timestamp != 0 || output_zero) {
```
Nakaner: After looking again at this code an my changes to it, I wonder if it wouldn't be better to return `1970-01-01T00:00:00Z` always if the timestamp is 0. We need something in the XML `timestamp` attribute. Or should OPL files don't have timestamps if they are 0.

joto:
Objects in the buffers are not `const`. If you don't declare them `const` you can change them. You can only make changes that don't change the size of the object, but that means clearing the metadata is fine. The only exception was the user name, but I just added a `clear_user()` function. (The reason there is no `set_user()` is that in the general case this could make the user name longer and there is no space in the buffer to do this. But clearing the user name always makes it shorter, so we can do that.)

So this should take care of the setting of the default values. Yes, it is a bit more complicated to use, but it is the cleaner approach. If I have special options on the output drivers, tomorrow the next one will come and say: I want an option to not write out tags, or whatever. It's really not the job of the output driver to mess with the data. (And yes, I should have thought of this earlier.)

Then there is the question of whether default values should be written to XML files or just not written out at all. I don't think we need to write these out. Every self-respecting XML reader already has to handle the case where this data is missing. This is a bit different for the PBF reader where there was clearly an expectation that all the data would be there or none. I am not totally sure about this and might be convinced otherwise though. But the first issue is with all this other software, you should first make an effort to get it up to date so that it works with your new data in current versions. If we add any other stuff to libosmium than only with a clear deprecation warning, so that we can get rid of it later when people had a chance to update their software to we don't have to carry around any hacks forever.

Nakaner:
Thank you for your response.

You prefer me abandoning this branch and adding this feature to the Cat command of Osmium Tool?

> Then there is the question of whether default values should be written to XML files or just not written out at all. I don't think we need to write these out. Every self-respecting XML reader already has to handle the case where this data is missing.

I doubt that there is any good XML reader in the OSM ecosystem. Even Osmosis which I expected to be quite good requires timestamps and version fields in the XML input.

> This is a bit different for the PBF reader where there was clearly an expectation that all the data would be there or none. I am not totally sure about this and might be convinced otherwise though. But the first issue is with all this other software, you should first make an effort to get it up to date so that it works with your new data in current versions. If we add any other stuff to libosmium than only with a clear deprecation warning, so that we can get rid of it later when people had a chance to update their software to we don't have to carry around any hacks forever.

It is the question how disruptive upcoming changes should be. The switch HTTPS only (redirects from port 80 to port 443) was announced multiple times but many users still don't follow redirects or send requests to port 443 directly.

Screenshot of the Munin graph (initially made for Wochennotiz), current graphs at [https://munin.openstreetmap.org/openstreetmap/ironbelly.openstreetmap/apache_accesses.html](https://munin.openstreetmap.org/openstreetmap/ironbelly.openstreetmap/apache_accesses.html)

ironbelly Munin log

Shall we move the discussion how disruptive upcoming changes should be to the Dev mailing list?

joto:
I have changed the XML writer to output the 0 timestamp correctly. As mentioned I don't think the approach using `use_default_metadata_values` is the best one here, so I am closing this PR.

PR #256, opened 2018-05-21
switch to HTTPS

wosch:
Most external sites now support HTTPS


joto:
Please do not change any files that are not created here but included from some upstream source (Doxygen config for instance.)

PR #266, opened 2018-08-24
Add unit test checking dumping and loading of dense indexes

Nakaner:
The test is inspired by test_file_based_index.

### DIFF
--- a/test/t/index/test_dump_and_load_index.cpp
+++ b/test/t/index/test_dump_and_load_index.cpp
@@ -1,0 +1,45 @@
+#include "catch.hpp"
+
+#include <osmium/index/detail/tmpfile.hpp>
+#include <osmium/index/map/dense_file_array.hpp>
+#include <osmium/index/map/dense_mmap_array.hpp>
+#include <osmium/osm/location.hpp>
+#include <osmium/osm/types.hpp>
+
+using dense_mmap_array = osmium::index::map::DenseMmapArray<osmium::unsigned_object_id_type, osmium::Location>;
+using dense_file_array = osmium::index::map::DenseFileArray<osmium::unsigned_object_id_type, osmium::Location>;
+
+TEST_CASE("Dump DenseMmapArray, load as DenseFileArray") {
+    const int fd = osmium::detail::create_tmp_file();
+    REQUIRE(osmium::file_size(fd) == 0);
+    constexpr const size_t MIN_INDEX_SIZE = sizeof(dense_file_array::element_type);
+
+    const osmium::unsigned_object_id_type id1 = 12;
+    const osmium::unsigned_object_id_type id2 = 3;
+    const osmium::unsigned_object_id_type id3 = 7;
+    const osmium::Location loc1{1.2, 4.5};
+    const osmium::Location loc2{3.5, -7.2};
+    const osmium::Location loc3{-12.7, 14.5};
+
+    dense_mmap_array index;
+    index.set(id1, loc1);
+    index.set(id2, loc2);
+    index.set(id3, loc3);
+
+    // We don't sort the index because sorting is not required for DenseMmapArray.
+    // dump to file
+    index.dump_as_array(fd);
+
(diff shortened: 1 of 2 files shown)


test/t/index/test_dump_and_load_index.cpp
```diff
@@ -0,0 +1,45 @@
...
+    const osmium::Location loc3{-12.7, 14.5};
+
+    dense_mmap_array_t index;
+    index.set(id1, loc1);
+    index.set(id2, loc2);
+    index.set(id3, loc3);
+
+    // We don't sort the index because sorting is not required for DenseMmapArray.
+    // dump to file
+    REQUIRE_NOTHROW(index.dump_as_array(fd));
```
joto: Not sure what the `REQUIRE_NOTHROW` helps here. If it throws the program will crash, so the test will fail.

test/t/index/test_dump_and_load_index.cpp
```diff
@@ -0,0 +1,45 @@
+#include "catch.hpp"
+
+#include <osmium/index/detail/tmpfile.hpp>
+#include <osmium/index/map/dense_file_array.hpp>
+#include <osmium/index/map/dense_mmap_array.hpp>
+#include <osmium/osm/location.hpp>
+#include <osmium/osm/types.hpp>
+
+using dense_mmap_array_t = osmium::index::map::DenseMmapArray<osmium::unsigned_object_id_type, osmium::Location>;
```
joto: Please don't use types with `_t` names. I know this is quite common but strictly speaking all names ending in `_t` are reserved by the C++ standard (even those in other names spaces).

test/t/index/test_dump_and_load_index.cpp
```diff
@@ -0,0 +1,45 @@
...
+using dense_file_array_t = osmium::index::map::DenseFileArray<osmium::unsigned_object_id_type, osmium::Location>;
+
+TEST_CASE("Dump DenseMmapArray, load as DenseFileArray") {
+    const int fd = osmium::detail::create_tmp_file();
+    REQUIRE(osmium::file_size(fd) == 0);
+    constexpr const size_t S = sizeof(dense_file_array_t::element_type);
```
joto: `S` is too short a name.

test/t/index/test_dump_and_load_index.cpp
```diff
@@ -0,0 +1,45 @@
...
+    dense_mmap_array_t index;
+    index.set(id1, loc1);
+    index.set(id2, loc2);
+    index.set(id3, loc3);
+
+    // We don't sort the index because sorting is not required for DenseMmapArray.
+    // dump to file
+    REQUIRE_NOTHROW(index.dump_as_array(fd));
+
+    REQUIRE(osmium::file_size(fd) >= (6 * S));
```
joto: Why is this `6 * S`.  Shouldn't it be `3 * S` ?

test/t/index/test_dump_and_load_index.cpp
Nakaner: removed

test/t/index/test_dump_and_load_index.cpp
Nakaner: `_t` removed

test/t/index/test_dump_and_load_index.cpp
Nakaner: I chose a longer name in uppercase (because it's a constant) but please have a look at https://github.com/osmcode/libosmium/blob/master/test/t/index/test_file_based_index.cpp#L24 ;)

test/t/index/test_dump_and_load_index.cpp
Nakaner: Yes, you are right. I fixed it.

Nakaner:
I added a test for DenseMemArray and put the DenseMmapArray test into `#ifdef __linux__` because DenseMmapArray does only work on Linux.

PR #275, opened 2018-12-12
Add test for Map::dump_as_list

Nakaner:
PR #266 added the test for `Map::dump_as_array`, this PR adds the same for `dump_as_list`.


joto:
Thanks. Could you use lambdas instead of `std::bind`? Bind is rather magic with those placeholders and isn't needed any more now that we have propert lambdas.

Nakaner:
I will replace `std::bind` by lambdas tomorrow.

Nakaner:
done

PR #281, opened 2019-06-26
Fix error: comparison of integer expression of different signedness

DennisOSRM:
Fixes the following issue:

```bash
error: comparison of integer expressions of different signedness: ‘std::size_t’ {aka ‘long unsigned int’} and ‘long int’ [-Werror=sign-compare]
  184 |             assert(std::numeric_limits<off_t>::max() >= std::numeric_limits<size_t>::max() || new_size <= std::numeric_limits<off_t>::max());
      |                                                                                               ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/include/osmium/util/file.hpp
+++ b/include/osmium/util/file.hpp
@@ -181,7 +181,7 @@ namespace osmium {
             // https://msdn.microsoft.com/en-us/library/whx354w1.aspx
             if (::_chsize_s(fd, static_cast<__int64>(new_size)) != 0) {
 #else
-            assert(std::numeric_limits<off_t>::max() >= std::numeric_limits<size_t>::max() || new_size <= std::numeric_limits<off_t>::max());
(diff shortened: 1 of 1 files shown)


DennisOSRM:
cc @joto

DennisOSRM:
friendly ping 😁

joto:
Sorry, I have seen this, but I believe your fix isn't correct. This is something I have to look at when I have more time and a clear head.

DennisOSRM:
> Sorry, I have seen this, but I believe your fix isn't correct. This is something I have to look at when I have more time and a clear head.

Thanks for the feedback, looking forward.

joto:
I believe your patch doesn't work correctly because if `off_t` can store something larger than `size_t` (which is what the first check in the `assert` is checking for), then the `static_cast` will do a narrowing conversion and things will break. I have checked now on 64bit and 32bit systems and I believe the assert here isn't necessary, because `off_t` is either the same size or larger than `size_t` (modulo signed vs. unsigned, sigh). So I have removed the assert entirely.

Issue #277, opened 2019-03-06
Error when building against proj 6.0.0


nextstopsun:
I'm trying to build libosmium with latest proj library.
I'm getting a test error:
```
Scanning dependencies of target geom_test_factory_with_projection
[ 33%] Building CXX object test/CMakeFiles/thread_test_util.dir/t/thread/test_util.cpp.o
[ 33%] Building CXX object test/CMakeFiles/geom_test_factory_with_projection.dir/t/geom/test_factory_with_projection.cpp.o
[ 34%] Linking CXX executable util_test_misc
In file included from /usr/src/libosmium/include/osmium/geom/projection.hpp:50:0,
                 from /usr/src/libosmium/test/t/geom/test_factory_with_projection.cpp:4:
/usr/local/include/proj_api.h:37:2: error: #error 'To use the proj_api.h you must define the macro ACCEPT_USE_OF_DEPRECATED_PROJ_API_H'
 #error 'To use the proj_api.h you must define the macro ACCEPT_USE_OF_DEPRECATED_PROJ_API_H'
```

What needs to be done to build successfully? 
Any special configure keys?

nextstopsun:
[This](https://github.com/OSGeo/proj.4/blob/6.0/src/proj_api.h#L37) where this error comes from.
I guess other headers should be used instead.

Nakaner:
Well, you use Proj 6 which has [major changes](https://github.com/OSGeo/proj.4/releases/tag/6.0.0) to the API.

joto:
As @Nakaner said, there seems to be major changes in version 6. It seems we could just set `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` before doing the include, but I'd rather use the new API correctly from the beginning. If course this would need some `#ifdef`s or so, so that it still works with the old library version. If somebody wants to figure this out and send a pull request, that'd be great.

cyper85:
> 
> 
> As @Nakaner said, there seems to be major changes in version 6. It seems we could just set `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` before doing the include, but I'd rather use the new API correctly from the beginning. If course this would need some `#ifdef`s or so, so that it still works with the old library version. If somebody wants to figure this out and send a pull request, that'd be great.

This isn't the solution. [proj_api.h will be removed in proj v7](https://github.com/OSGeo/PROJ/issues/836). The macro is only a workaround for v6. The best solution is to disintegrate the old api.

[...]

PR #284, opened 2019-07-18
Changing proj API from v4 to v6

cyper85:
https://github.com/osmcode/libosmium/issues/277

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -177,7 +177,7 @@ install:
   - git clone --quiet --depth 1 https://github.com/mapbox/protozero.git ../protozero
   - |
     if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
-      brew remove gdal
+      brew remove --ignore-dependencies gdal
       brew install cmake boost google-sparsehash gdal proj || true
     fi
   - cmake --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@
 
 language: generic
 
-dist: xenial
+dist: bionic
 
 #-----------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


cyper85:
https://git.neumann-andreas.de/andreas/libosmium-docker/blob/master/Dockerfile

Test-Docker-Build-Container with alpine:latest and proj v6

joto:
Thanks for the pull request. Please check what happened that all lines show up as changed. Also for backwards compatibility the old API must still work, please add some kind of ifdef-magic or so to check for that.

Nakaner:
> Also for backwards compatibility the old API must still work, please add some kind of ifdef-magic or so to check for that.

@cyper85 You might have a look at https://github.com/Nakaner/osmium-geos-factory/blob/master/include/osmium_geos_factory/geos_factory.hpp#L47 where I deal with changes in GEOS's C++ API. `proj.h` defines the macros `PROJ_VERSION_MAJOR`, `PROJ_VERSION_MINOR` and `PROJ_VERSION_MAJOR` patch for that purpose.

joto:
* It still shows all lines as changed. Is there a problem with line endings?
* `__has_include` is only available since C++17, but libosmium should work from C++11 on.

cyper85:
Okay,

I had to modify .trafis.yml. Due the little Change I add osx10.2 and osx11 to the installation.

In osx will build proj in v6.1, so you can see in your tests, that it works.

joto:
We have a temporary solution (see #277) now.

This patch is problematic on several levels. Some problems I have mentioned above. Then there is the use of `const_cast`. And even if we solve all this, the API that libosmium provides just can't be implemented suitably with the new proj API. We need a better solution.

PR #285, opened 2019-08-13
AppVeyorCI: Fix MSYS2 build

tusharpm:
The fix for gdal is available upstream and references the correct json-c dll.
The workaround (until reported/fixed upstream) provides a spatialite dependency: readline 7.

### DIFF
--- a/build-msys2.bat
+++ b/build-msys2.bat
@@ -13,8 +13,8 @@ bash -lc "pacman -S --needed --noconfirm mingw-w64-x86_64-gcc mingw-w64-x86_64-g
 bash -lc "pacman -S --needed --noconfirm mingw-w64-x86_64-postgresql mingw-w64-x86_64-netcdf mingw-w64-x86_64-crypto++"
 call C:\msys64\mingw64\bin\gem.cmd install json
 
-REM Workaround for problem with gdal (see https://github.com/osmcode/libosmium/issues/262)
-copy /y C:\msys64\mingw64\bin\libjson-c-4.dll C:\msys64\mingw64\bin\libjson-c-3.dll
+REM Workaround for problem with spatialite (see https://github.com/osmcode/libosmium/issues/262)
+copy /y C:\msys64\mingw64\bin\libreadline8.dll C:\msys64\mingw64\bin\libreadline7.dll
 
 echo "Setting PROJ_LIB variable for correct PROJ.4 working"
 set PROJ_LIB=c:\msys64\mingw64\share\proj


joto:
I am not sure I understand. The MSYS2 build just breaks off at some point. How is this related to readline?

tusharpm:
The current version of [readline](https://packages.msys2.org/package/mingw-w64-x86_64-readline) is 8 since March, 2019, which is used by default to install.
The most recent version for [spatialite-tools](https://packages.msys2.org/package/mingw-w64-x86_64-spatialite-tools) is from January, 2017. It depends on readline:7 [which is not mentioned in its PKGBUILD](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-spatialite-tools/PKGBUILD).
The upstream fix needs to add readline and build spatialite-tools with the new version.
Until then, the new version of readline:8 can be used as a placeholder for readline:7.

joto:
Okay, I think I see where the confusion is. The `build-msys2.bat` file is not a general build-on-msys2 script, it is specifically there to do the CI build on Travis. And while the build might be broken on your system and can be fixed this way, the Travis build isn't broken. Well, it breaks off at some point, but that breakage doesn't seem to be related to your fix. So there is no need here to change the bat file.

tusharpm:
Actually, `build-msys2.bat` is used for the Windows - MSYS2 build on AppVeyor CI.
It can be used for building locally, but the issue is visible on CI as well.
README.md currently shows a red badge: [Appveyor Build status](https://ci.appveyor.com/project/lonvia/libosmium/branch/master). The build log can be seen [here](https://ci.appveyor.com/project/lonvia/libosmium/builds/26614949/job/fp2223l8sx9vw31o).
This fix (rather, workaround) can get it to green.
Once fixed,  it might be worthwhile to integrate AppVeyor for checking PRs on Windows as well.

joto:
Sorry, I meant "Appveyor" when I said "travis". Where are you seeing that this fix, fixes the Appveyor build? I have no idea why Appveyor doesn't build the PR, travis does. Do you know how this can be changed?

tusharpm:
I'm sorry, I deleted my AppVeyor project history after testing. I've triggered [a new build now](https://ci.appveyor.com/project/tusharpm/libosmium/builds/26692715).

> Do you know how this can be changed?

I think the option `Do not build on "Pull request" events` should be unchecked on [the repository's AppVeyor settings page](https://ci.appveyor.com/project/lonvia/libosmium/settings). There might be other configuration options to fix though.

lonvia:
I've just checked the Appveyor settings and they look fine. May be it wasn't built because @tusharpm  did build it on his form already? There is no button either to force a rebuild for a PR. Has to be done via the API (which I'm not exactly motivated to learn to use).

tusharpm:
All [PRs](https://github.com/osmcode/libosmium/pulls?q=is%3Apr) (including this one) show only the Travis CI status (usually success) for the tip of the branch. I had deleted my AppVeyor project to avoid triggering the system twice for the same commit.
If the integration works properly, adding a commit (or force pushing) to my branch should trigger the build for the PR. Please let me know if you want to try that.

PR #292, opened 2019-12-08
index speedup: after mmap, call madvise with MADV_RANDOM

bdon:
This change results in a small but noticeable improvement of about 8% on my program which uses SparseFileArray for indexing node locations. Runs on a ~6GB .osm.pbf go from ~3450 seconds to 3200 seconds. 

http://man7.org/linux/man-pages/man2/madvise.2.html it should not change the behavior of the program but tweak the kernel paging slightly. Node locations seem random enough that readahead will always be detrimental, not sure where else this class is used though.

If others can reproduce the gains here and think it's useful, I can make sure it works on other platforms - only tested on Linux so far.

### DIFF
--- a/include/osmium/util/memory_mapping.hpp
+++ b/include/osmium/util/memory_mapping.hpp
@@ -573,6 +573,7 @@ inline osmium::util::MemoryMapping::MemoryMapping(std::size_t size, mapping_mode
     m_fd(resize_fd(fd)),
     m_mapping_mode(mode),
     m_addr(::mmap(nullptr, m_size, get_protection(), get_flags(), m_fd, m_offset)) {
+    madvise(m_addr,m_size,MADV_RANDOM);
     assert(!(fd == -1 && mode == mapping_mode::readonly));
     if (!is_valid()) {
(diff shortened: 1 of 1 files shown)


joto:
This is definitely not something that should be forced in the low-level `memory_mapping.hpp` code. There are different use cases that will show different access patterns. Even the most common use case has exactly those two patterns: While reading the nodes and creating the index, nodes come in order and pages are used in order. Then later, while adding the node locations to the ways, the access will be more or less random. We have to make sure we are not pessimizing anything here.
So we'd need a lot more tests to get an idea how performance looks in different cases and where to best integrate such code. In theory it should be possible to use different `madvise` settings for different parts of the program for the same piece of memory. Would be interesting to see some further analysis here.

And, as you already mentioned `madvise()` is a special thing in Linux, at a minimum we have to put some macro checks around it or so, so that the code still compiles in other systems. And, btw., please use the formatting conventions established in the surrounding code, in this case spaces after commas.

bdon:
OK, thanks for looking at it - I'll find a way to move `madvise` to my application code instead.

PR #303, opened 2020-04-10
Fix location attr variable name typo

CloudNiner:
Looks like the names for these arguments are swapped, which could lead to confusion and subtle bugs.


joto:
Good catch. Thank you!

PR #304, opened 2020-04-13
fix parsing lat bug from pbf densenode

zhaitianduo:
When parsing lat from pbf densenode,
I noticed lat is also parsed using convert_pbf_coordinate which also using m_lon_offset.

Per https://wiki.openstreetmap.org/wiki/PBF_Format
`latitude = .000000001 * (lat_offset + (granularity * lat))
longitude = .000000001 * (lon_offset + (granularity * lon))`

latitude should be calculated from lat_offset

### DIFF
--- a/include/osmium/io/detail/pbf_decoder.hpp
+++ b/include/osmium/io/detail/pbf_decoder.hpp
@@ -260,8 +260,12 @@ namespace osmium {
                     }
                 }
 
-                int32_t convert_pbf_coordinate(const int64_t c) const noexcept {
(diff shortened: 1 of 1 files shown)


zhaitianduo:
@joto could help take a look?

joto:
Good catch. Thanks!

I have never seen anybody use anything but 0 offsets or non-default granularity...

zhaitianduo:
Thanks @joto 
I wrote a pbf writer by myself using Golang, I tried to verify the output of the writer using osmium and found this issue.

PR #326, opened 2021-03-02
match_all for tag filter list

Zverik:
Adds a function to the tags filter that tests that all the rules have a match in a tag list. I don't know STL well enough to employ it in the function. It is used in https://github.com/osmcode/osmium-tool/pull/209

Feel free to rename and rewrite it better, or to close the PR if the function does not fit.

### DIFF
--- a/include/osmium/tags/tags_filter.hpp
+++ b/include/osmium/tags/tags_filter.hpp
@@ -135,6 +135,32 @@ namespace osmium {
         }
 
         /**
+         * Testing that all rules match for a set of tags.
+         *
+         * @param tags A list of tags.
+         * @returns False if the list of tags is empty,
+         *          or some rules are not matched.
+         */
+        bool match_all(const osmium::TagList& tags) const noexcept {
+            if (tags.empty() || m_rules.empty()) {
+                return false;
+            }
+            for (const auto& rule : m_rules) {
+                bool found = false;
+                for (const auto& tag : tags) {
+                    if (rule.second(tag)) {
+                        found = true;
+                        break;
+                    }
+                }
+                if (!found) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        /**
          * Return the number of rules in this filter.
          *
          * Complexity: Constant.


joto:
The TagsFilter is designed in a way that this shouldn't be necessary. You can always define the rules in a way to make it possible to check for all rules to match. To do this invert the result values of the rules and then invert the overall result. Basically this is using [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

Zverik:
I see — I can probably do that. Thanks for the link.

Zverik:
Okay, I'm a bit lost here. `osmium::tags::match_any_of` function that's used in `command_tags_filter` runs `std::any_of` for a list of tags and a tag filter. Inside, each tag is passed to `TagsFilterBase::operator()`, which iterates over the rules. Hence we cannot determine whether _all_ rules match a set of tags, only when any of the rules matches. All rules cannot match all tags.

That was the reason for this pull request: to invert the iteration order, and find tags for each rule, not rules for each tag.

Is it possible to do this without modifying libosmium?

Zverik:
And I cannot put the `match_all` function outside of `TagsFilterBase`, because `m_rules` field is private with no getters (although for some reason there's `count()` method). It'd be simpler if it had e.g. `cbegin()` and `cend()` for rules.

joto:
Using the TagsFilter is not as easy as it maybe could be. But it is trying to be implemented as flexible building blocks that can be assembled in different ways. I myself get sometimes confused and don't use them that often... I have added an example `osmium_tags_filter.cpp` in the `examples` dir showing how it is to be used. The PR here mixes up the TagsFilter with the TagList in a way that feels overly specialized to me. And it isn't necessary, as can be seen in the example. Sorry that I only found the time now to look at this more closely.

Zverik:
It's okay — thanks Jochen for looking at the code! Since then I implemented the process that needed this operation differently using pure libosmium.Maybe somebody else would get an idea on how to proceed after looking at this PR.

PR #332, opened 2021-06-01
Explicitly export exception types.

oraluben:
This should fix osmcode/pyosmium#165

Ref: pybind/pybind11#2999, [GCC Wiki](https://gcc.gnu.org/wiki/Visibility)

### DIFF
--- a/include/gdalcpp.hpp
+++ b/include/gdalcpp.hpp
@@ -42,6 +42,8 @@ DEALINGS IN THE SOFTWARE.
 #include <ogr_api.h>
 #include <ogrsf_frmts.h>
 
+#include <osmium/util/compatibility.hpp>
+
 #include <cstdint>
 #include <algorithm>
 #include <memory>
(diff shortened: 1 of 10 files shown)


oraluben:
Working on CI failure.

joto:
Looks to me like you probably didn't include the `compatibility.hpp` everywhere that's needed. For instance in `include/osmium/memory/buffer.hpp`.

oraluben:
> Looks to me like you probably didn't include the `compatibility.hpp` everywhere that's needed. For instance in `include/osmium/memory/buffer.hpp`.

Thanks for your reply! They already included the file (e.g. [in `buffer.hpp`](https://github.com/osmcode/libosmium/blob/master/include/osmium/memory/buffer.hpp#L39)), and I also successfully compiled the project. Is there anything different I should notice between local and the CI's environment?

joto:
I have merged this now (in slightly different form). Sorry it took so long, but I needed to understand what's happening here first.

joto:
Because VS is generating warnings I wrapped the code defining `OSMIUM_EXPORT` into an `#ifdef OSMIUM_DEFINE_EXPORT`. So you have to compile with `-DOSMIUM_DEFINE_EXPORT` if you need this functionality. Can you provide a PR for Pyosmium if that works for you?

oraluben:
@joto thanks! I'll try to draft a PR there.

PR #334, opened 2021-07-26
Read PBF files without an extra thread

joto:
Unlike XML or OPL files, PBF files are never packed into gzip or bzip2
as a whole. So we don't need the extra thread doing that unpacking.
This speeds up reading of PBF files.


joto:
Rebased version of this has been merged.

PR #343, opened 2021-12-02
Added SparseMemCompactArray implementation and tests

vlopezferrando:
Hello, first of all, thanks for developing libosmium, I've been a user for some years and it is an exceptional piece of software!

Lately, I have been using libosmium in different applications that required the indexing of node locations, and I came up with an idea to make indexing faster. Here I provide a pull request that implements this idea. Please, could you review it, give any feedback, and eventually merge it in libosmium?

## The idea

I observed the sparse memory array index is based on a vector of pairs `(id, loc)`, such as:

```
index: (1, {1.0, 1.0}), (2, {2.0, 2.0}), (3, {3.0, 3.0}), (8, {8.0, 8.0}), (9, {9.0, 9.0})
```

However, it is very common in OpenStreetMap to find sets of sequential ids, even in small files (because they may have been added together as part of the same way).

This property allows to separately store locations and ids, storing only the ids that are not immediately after the previous one. In the case of the previous example:

```
index: (1, 0), (8, 3)
locations: {1.0, 1.0}, {2.0, 2.0}, {3.0, 3.0}, {8.0, 8.0}, {9.0, 9.0}
```

In general, I have observed that the number of sequential chunks of ids is about 20% of the total number of ids.

## Implementation

I have implemented this index as a class named `SparseMemCompactArray`, though probably there is a better name for it.

This class includes a `sort` method, which needs to be run if ids have been inserted out of order.
This method is the most inefficient one memorywise, as it will need to build a whole vector of pairs (like `SparseMemArray`) and then insert all elements again.

## Test

I have added unit testing for this class, and in fact expanded the existing tests on `test_id_to_location`.
Instead of using only 2 values, I have created some vectors with tens of values, both ordered and unordered, and extensively checked for both added and missing values.

I have replaced the previous `test_func_real` test with the new one, but this could also be added as a separate test.

I removed the `ENABLE_IF ${SPARSEHASH_FOUND}` for `test_id_to_location` in `test/CMakeLists.txt` because otherwise I wasn't able to get the test to run. Probably I was missing something and this change should be reverted.

I also commented the test case `Map Id to location: SparseMemMap` because it didn't compile for some reason. This should also be fixed before merging.

Tests are passing:

```
ctest -R "index_test_id_to_location"
Test project /home/victor/git/libosmium/build
    Start 46: index_test_id_to_location
1/1 Test #46: index_test_id_to_location ........   Passed    0.07 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
fast     =   0.07 sec*proc (1 test)
index    =   0.07 sec*proc (1 test)
unit     =   0.07 sec*proc (1 test)

Total Test time (real) =   0.07 sec
```

## Benchmark

Finally, I have run the index map benchmark (`run_benchmark_index_map.sh`) for maps of different sizes downloaded from geofabrik.de, specifically: Europe (26 GB), Germany (2.9 GB), Belgium (465 MB), Iceland (52 MB).
These are the results:

```
BENCHMARK: index_map
---------------------
BUILD:
build type	: Dev
compiler	: /usr/lib/ccache/c++
CXX version	: c++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
CXX flags	: -O3 -g
---------------------
CPU:
model name	: AMD Ryzen 9 3900X 12-Core Processor
cpu MHz		: 3606.371
cpu cores	: 12
siblings	: 24
---------------------
MEMORY:
              total        used        free      shared  buff/cache   available
Mem:       65852236    10943252      643060      303464    54265924    54406508
Swap:             0           0           0
---------------------
RESULTS:
# file size num mem time cpu_kernel cpu_user cpu_percent cmd options

europe-latest.osm.pbf 26053108202 1 45029616 546.20 28.85 1587.04 295% osmium_benchmark_index_map europe-latest.osm.pbf sparse_mmap_array
europe-latest.osm.pbf 26053108202 2 45026796 542.16 29.06 1587.98 298% osmium_benchmark_index_map europe-latest.osm.pbf sparse_mmap_array
europe-latest.osm.pbf 26053108202 3 45029764 543.55 28.75 1591.40 298% osmium_benchmark_index_map europe-latest.osm.pbf sparse_mmap_array
europe-latest.osm.pbf 26053108202 1 38737268 378.02 32.41 1354.23 366% osmium_benchmark_index_map europe-latest.osm.pbf sparse_mem_compact_array
europe-latest.osm.pbf 26053108202 2 38738824 376.49 32.57 1351.70 367% osmium_benchmark_index_map europe-latest.osm.pbf sparse_mem_compact_array
europe-latest.osm.pbf 26053108202 3 38739196 375.63 32.43 1350.82 368% osmium_benchmark_index_map europe-latest.osm.pbf sparse_mem_compact_array

germany-latest.osm.pbf 3755159693 1 8431576 74.41 5.46 215.20 296% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mem_array
germany-latest.osm.pbf 3755159693 2 8431500 74.00 5.74 213.88 296% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mem_array
germany-latest.osm.pbf 3755159693 3 8432472 72.96 5.29 212.43 298% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mem_array
germany-latest.osm.pbf 3755159693 1 5749100 74.35 9.05 222.51 311% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mmap_array
germany-latest.osm.pbf 3755159693 2 5731828 74.53 8.76 222.65 310% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mmap_array
germany-latest.osm.pbf 3755159693 3 5745884 74.61 8.69 222.43 309% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mmap_array
germany-latest.osm.pbf 3755159693 1 5750824 77.18 11.08 227.59 309% osmium_benchmark_index_map germany-latest.osm.pbf sparse_file_array
germany-latest.osm.pbf 3755159693 2 5740080 77.23 10.95 229.20 310% osmium_benchmark_index_map germany-latest.osm.pbf sparse_file_array
germany-latest.osm.pbf 3755159693 3 5735224 76.99 10.18 225.39 305% osmium_benchmark_index_map germany-latest.osm.pbf sparse_file_array
germany-latest.osm.pbf 3755159693 1 5410152 55.03 4.87 195.63 364% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mem_compact_array
germany-latest.osm.pbf 3755159693 2 5411168 55.00 5.01 193.58 361% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mem_compact_array
germany-latest.osm.pbf 3755159693 3 5410940 55.25 4.96 195.16 362% osmium_benchmark_index_map germany-latest.osm.pbf sparse_mem_compact_array

belgium-latest.osm.pbf 465174985 1 1075712 6.72 0.65 25.28 385% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mem_array
belgium-latest.osm.pbf 465174985 2 1075616 6.81 0.83 25.52 387% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mem_array
belgium-latest.osm.pbf 465174985 3 1077216 6.80 0.88 25.84 393% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mem_array
belgium-latest.osm.pbf 465174985 1 924292 6.92 1.84 25.97 401% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mmap_array
belgium-latest.osm.pbf 465174985 2 924208 6.88 1.97 25.88 404% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mmap_array
belgium-latest.osm.pbf 465174985 3 923392 6.92 1.93 25.53 396% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mmap_array
belgium-latest.osm.pbf 465174985 1 924660 7.51 2.24 26.68 385% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_file_array
belgium-latest.osm.pbf 465174985 2 924832 7.43 2.29 26.42 386% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_file_array
belgium-latest.osm.pbf 465174985 3 925028 7.38 2.23 26.36 387% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_file_array
belgium-latest.osm.pbf 465174985 1 690492 4.96 0.72 23.92 496% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mem_compact_array
belgium-latest.osm.pbf 465174985 2 689472 4.97 0.71 23.82 493% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mem_compact_array
belgium-latest.osm.pbf 465174985 3 694392 4.84 0.82 23.83 509% osmium_benchmark_index_map belgium-latest.osm.pbf sparse_mem_compact_array

iceland-latest.osm.pbf 52291816 1 313100 0.74 0.21 3.03 437% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mem_array
iceland-latest.osm.pbf 52291816 2 313220 0.74 0.20 2.92 420% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mem_array
iceland-latest.osm.pbf 52291816 3 313248 0.75 0.21 3.03 429% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mem_array
iceland-latest.osm.pbf 52291816 1 282076 0.74 0.31 2.95 441% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mmap_array
iceland-latest.osm.pbf 52291816 2 273020 0.73 0.29 2.95 440% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mmap_array
iceland-latest.osm.pbf 52291816 3 279664 0.73 0.29 2.95 440% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mmap_array
iceland-latest.osm.pbf 52291816 1 274844 0.80 0.30 3.05 416% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_file_array
iceland-latest.osm.pbf 52291816 2 276836 0.79 0.35 3.01 425% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_file_array
iceland-latest.osm.pbf 52291816 3 269616 0.78 0.31 2.99 423% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_file_array
iceland-latest.osm.pbf 52291816 1 220260 0.46 0.16 2.66 602% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mem_compact_array
iceland-latest.osm.pbf 52291816 2 217588 0.47 0.16 2.79 619% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mem_compact_array
iceland-latest.osm.pbf 52291816 3 213476 0.47 0.16 2.69 596% osmium_benchmark_index_map iceland-latest.osm.pbf sparse_mem_compact_array
```

For Europe I only compared `sparse_mem_compact_array` to `sparse_mmap_array`, as `sparse_mem_array` required too much memory, and `sparse_file_array` performs very similarly to `sparse_mmap_array`.

As it can be noted, the new `SparseMemCompactArray` index improves by 35% in terms of memory use and 25% in terms of speed compared to `SparseMemArray`, the index that is more similar.


vlopezferrando:
I have fixed a bug (adding id `0` when the index was empty resulted in a broken index).

I have also added tests for adding very high ID values (using `numeric_limits`), and have applied them to all Sparse indexes.

vlopezferrando:
I have replaced `or` and `and` for `||` and `&&`, as it seems this broke windows compilation.

vlopezferrando:
I have also removed the `id = -1` which caused an error in the unit testing:
```
 /__w/libosmium/libosmium/include/osmium/index/map/sparse_mem_compact_array.hpp:68:33: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'unsigned long' changed the value to 18446744073709551615 (64-bit, unsigned)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /__w/libosmium/libosmium/include/osmium/index/map/sparse_mem_compact_array.hpp:68:33 in 
```

@joto, can you run the pipeline again, as I think all issues have been addressed? Thanks.

joto:
This is a promising idea and well executed and put into the PR, thanks! I have to look at this in more detail, but see these issues up front:

1. the include guard (`#ifndef OSMIUM_INDEX_MAP_SPARSE_MEM_MAP_HPP`...) is wrong. This should be based on the file name. This is most likely the reason that you have problems with some tests.
2. The `dump_as_list()` function is based on `sorted_vector_of_pairs()` which means this will need quite a lot of memory. It should be possible to implement this in a different way, writing out blocks of, say, 1000 entries by iterating over the index and elements at the same time.
3. Due to the extra memory needed sorting is basically out of the question for most use cases. I wonder if we can find some trick to do the sorting without it, or maybe do it piece by piece so the memory consuption will be kept in check.

vlopezferrando:
@joto I just fixed the remaining failing test in the pipeline. I fixed the include guard and recovered the commented test: it works now :).

Regarding a more efficient sorting of the index, if I understand correctly your suggestion in point 2, the idea would be: build a vector of pairs <id, location> while shrinking the size of the locations and index vector while you iterate over them. In the end, you have a single vector<id, location>. Then you can sort this vector, and then add all the elements back to the index, also while reducing its size. It will be as slow as the current implementation, but using about half the memory, so I agree it would be much better.

On whether there is a better solution, I thought the following: the problem with sorting the index is that we lose the order of the locations, and then it is impossible to know how many contiguous ids there are in the locations index. We could solve this problem by adding all the <id, location_i> pair to the index. For example, given the following table:

```
locations = 3, 4, 2, 0, 1
index = (3, 0), (2, 2), (0, 3)
```
We could add the test of the locations to the index:
```
locations = 3, 4, 2, 0, 1
index = (3, 0), (2, 2), (0, 3), (4, 1), (1, 4)
```
Now we can sort:
```
locations = 3, 4, 2, 0, 1
index = (0, 3), (1, 4), (2, 2), (3, 0), (4, 1)
```
And finally we can erase from the vector all the locations that have both sequential IDs and sequential location indices:
```
locations = 3, 4, 2, 0, 1
index = (0, 3), (2, 2), (3, 0)
```
If we followed this algorithm, the max amount of memory required would be `nlocations x 24 bytes` (8 bytes for the location, and 8+8 bytes for the pair <ID, location_i>).

There is probably a better way, but this is what I have come up for now.

joto:
> Regarding a more efficient sorting of the index, if I understand correctly your suggestion in point 2, the idea would be: build a vector of pairs <id, location> while shrinking the size of the locations and index vector while you iterate over them.

No, that's not going to work. Shrinking a vector either does nothing (if you just `resize()`) or it creates a temporary copy (if you use `shrink_to_fit()` which means you are using about twice the memory for some time. So either you use the existing memory in the vector in some different way (but I don't know exactly what that would be) or you do the sorting piece by piece in some way. Maybe sort blocks of the data and then merge the blocks in some way or so.

Come to think of it, we'll always have the vector resizing problem with this implementation. I.e. we need twice the amount of memory when the vectors resize. This can be avoided by putting the actual data into blocks that never resize for the price of another indirection. Somewhat like what I have done in the [node location store in osm2pgsql](https://github.com/openstreetmap/osm2pgsql/blob/master/src/node-locations.hpp). That implementation is very memory-efficient, but it doesn't allow updates and can't do sorting.

joto:
This is PR seems dead. Sorry, but I can't merge it the way it is. Closing.

PR #349, opened 2022-03-18
CheckOrder handler with configurable duplicate handling strategy

MateuszGrabka-TomTom:
CheckOrder handler throws an exception when the same ID appears twice in the input. Currently, it is impossible to print all duplicates if needed. This pull request introduces possibility to configure handler's behaviour to print duplicates instead of thrownig exception.

Backward compatibility is preserved. By default, old behaviour is preserved (throwing exception).

### DIFF
--- a/include/osmium/handler/check_order.hpp
+++ b/include/osmium/handler/check_order.hpp
@@ -43,6 +43,7 @@ DEALINGS IN THE SOFTWARE.
 
 #include <stdexcept>
 #include <string>
+#include <iostream>
 
 namespace osmium {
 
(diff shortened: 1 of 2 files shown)


joto:
I am not sure I like printing out stuff in a library, especially not just to `cout`. Libraries should really not do IO if that can be avoided. What if the application wants to write this data to a file?

Can you tell me something about your use case? Why is this something that needs to be solved in the library? That check isn't really all that complicated, so maybe that's something the application can just do itself?

MateuszGrabka-TomTom:
> 

I know printing to `cout` should be avoided if possible but this is exactly what I need. That is why I've added a `duplicate_handling` enum there. If you don't want to print it, it is still fine, you can use THROW_EXCEPTION. If there will be a need in the future to have other reaction, additional enum can be added (DO_NOTHING as an example) or even next parameter can be added to change cout to file output (I don't need it now so I did not want to overdesing it).

My ultimate goal is to extend `check-refs` from `osmium-tool`. It is using CheckOrder handler internally. Currently, `check-refs` crashes when duplicated IDs are detected. I would like to print out all referential integerity and duplicated ids issues from input PBF ultimately.

joto:
I am sorry, but I wont accept the PR in this form. I could imagine adding some more modular feature where the code run, when a problem is found, is plugged in using some kind of policy class. But hardcoding the output to `cout` isn't a good idea. It does come with a cost (compile time because of added include header, but mostly maintainence and testing needs for the added code) while only delivering a tiny benefit in a very special use case. I strive to offer a library of building blocks to mix and match code to build powerful tools with it, and this is not the way to go in my opinion.

Issue #353, opened 2022-07-26
io_test_bzip2 fails on Alpine Linux


brawer:
Test `io_test_bzip2` fails when building libosmum v2.18.0 on Alpine Linux 3.16.1.

The Alpine Linux distribution is commonly used in container-based cloud environments because its container images are much smaller than those of Debian or Ubuntu. Of course nobody would want to parse OSM files on embedded systems, but Alpine-based containers are often used for cloud servers and indexing pipelines.

## How to reproduce

1. Put the following content into a file called `Dockerfile`.
2. Run `docker build -t bugreport .` on the command line. I’m using a MacBook, but any other host system with Docker support should work as well. If you don’t like Docker, [Buildah](https://buildah.io/) is a free-as-in-speech alternative, but I haven’t tried.

```Dockerfile
FROM alpine:3.16.1

RUN apk add --no-cache build-base cmake git \
  boost boost-dev \
  bzip2 bzip2-dev \
  expat expat-dev \
  lz4 lz4-dev

WORKDIR /build

RUN git clone https://github.com/mapbox/protozero --single-branch --branch v1.7.1 && cd /build/protozero && mkdir build && cd build && cmake .. && make && make test && make install && rm -rf /build/protozero

RUN git clone https://github.com/osmcode/libosmium --single-branch --branch v2.18.0 && cd /build/libosmium && mkdir build && cd build && cmake .. && make

RUN cd /build/libosmium/build && make test

CMD /bin/sh
```

## Expected

All tests should pass.

## Observed

Test case `io_test_bzip2` is failing. For more details, comment out the second-last line in the above `Dockerfile` (the one that calls `make test`), build the container again, and then log into the virtual machine with `docker run -it bugreport`. When doing this and running the failing test case manually, I’m getting the following output:

```
# cd /build/libosmium/test

/build/libosmium/test # /build/libosmium/build/test/io_test_bzip2

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
io_test_bzip2 is a Catch v2.13.8 host application.
Run with -? for options

-------------------------------------------------------------------------------
Invalid file descriptor of bzip2-compressed file
-------------------------------------------------------------------------------
/build/libosmium/test/t/io/test_bzip2.cpp:10
...............................................................................

/build/libosmium/test/t/io/test_bzip2.cpp:11: FAILED:
[...]

PR #354, opened 2022-08-04
Check for bad file descriptors when opening bzip2-compressed files

brawer:
After this change, all tests pass on Alpine Linux.
Fixes https://github.com/osmcode/libosmium/issues/353.

### DIFF
--- a/include/osmium/io/bzip2_compression.hpp
+++ b/include/osmium/io/bzip2_compression.hpp
@@ -58,6 +58,8 @@ DEALINGS IN THE SOFTWARE.
 #include <string>
 #include <system_error>
 
+#include <fcntl.h>
+
 #ifndef _MSC_VER
 # include <unistd.h>
 #endif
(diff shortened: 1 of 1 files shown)


brawer:
The AppVeyor build failure seems unrelated to the change.
https://github.com/osmcode/libosmium/issues/355

joto:
Fails on Windows builds.

brawer:
> Fails on Windows builds.

@joto, mind approving running GitHub workflows for this change? Should be fixed on Windows, but as a first-time committer to this repo, I can’t start the workflow myself.

joto:
I am not too happy about that solution. Having to fix a minor problem that will never show up in real use on a fringe system is bad enough and now we need a fix for the fix (ie the ifdef) because it doesn't work on Windows. Maybe we need a different approach here. Presumably using a broken file descriptor with musl will still fail, but instead of in the fdopen somewhere down the line when we try to read some data. Maybe the test should be rewritten to accept failure then.

brawer:
Agree. See https://github.com/osmcode/libosmium/pull/357 for an alternative fix.

PR #357, opened 2022-08-08
Make bzip2 unit tests pass on musl-based systems

brawer:
Other than glibc, the musl standard C library does not check the
validity of file descriptors passed to `fdopen()`, so the constructor
of the (de)compressor does not throw an excption on musl-based systems.
Rather, an exception is thrown later, when the opened file is being used.
According to the musl team, the current musl behavior conforms to the POSIX
specification, and application code should be made agnostic to whether
or not the standard C library happens to validate file descriptors
passed to `fdopen()`.

After this change, all libosmium tests (including those for bzip2) are
passing on Alpine Linux, which uses musl as its standard C library.

Fixes https://github.com/osmcode/libosmium/issues/353.

### DIFF
--- a/test/t/io/test_bzip2.cpp
+++ b/test/t/io/test_bzip2.cpp
@@ -7,13 +7,19 @@
 
 #include <string>
 
+static void read_from_decompressor(int fd) {
+    osmium::io::Bzip2Decompressor decomp{fd};
+    decomp.read();
+    decomp.close();
+}
+
 TEST_CASE("Invalid file descriptor of bzip2-compressed file") {
-    REQUIRE_THROWS_AS(osmium::io::Bzip2Decompressor{-1}, std::system_error);
+    REQUIRE_THROWS(read_from_decompressor(-1));
 }
 
 TEST_CASE("Non-open file descriptor of bzip2-compressed file") {
     // 12345 is just a random file descriptor that should not be open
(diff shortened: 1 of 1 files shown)


test/t/io/test_bzip2.cpp
```diff
@@ -7,13 +7,19 @@
 
 #include <string>
 
+static void read_from_decompressor(int fd) {
+    osmium::io::Bzip2Decompressor decomp(fd);
```
joto, line 11: Please use {} for constructors. Same issue below.

test/t/io/test_bzip2.cpp
```diff
@@ -7,13 +7,19 @@
...
+    decomp.read();
+    decomp.close();
+}
+
 TEST_CASE("Invalid file descriptor of bzip2-compressed file") {
-    REQUIRE_THROWS_AS(osmium::io::Bzip2Decompressor{-1}, std::system_error);
+    REQUIRE_THROWS_AS(read_from_decompressor(-1), std::exception);
```
joto, line 17: If you don't check for a specific exception, just using `REQUIRE_THROWS()` is better. Same issue below.

joto (changes requested):
Looks much better to me now.

joto:
The Ubuntu 22.04 build error is unrelated. I just fixed that in master (I hope).

brawer:
Done. Many thanks for the review; please take another look.

PR #360, opened 2023-01-10
catch.hpp: restore PPC support

barracuda156:
PPC define taken from earlier version of `libosmium`: https://github.com/osmcode/libosmium/blob/b293e96dbe2e3a9afe400768a990e107f8af3f02/test/catch/catch.hpp#L2128

Second commit is non-functional: removal of few unneeded double spaces.

### DIFF
--- a/test/catch/catch.hpp
+++ b/test/catch/catch.hpp
@@ -7950,6 +7950,9 @@ namespace Catch {
         #define CATCH_TRAP() __asm__("int $3\n" : : ) /* NOLINT */
     #elif defined(__aarch64__)
         #define CATCH_TRAP()  __asm__(".inst 0xd4200000")
+    #elif defined(__POWERPC__)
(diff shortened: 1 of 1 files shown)


joto:
catch.hpp is included as is from its upstream source. It is not maintained here. Please solve this issue with upstream (https://github.com/catchorg/Catch2).

barracuda156:
> catch.hpp is included as is from its upstream source. It is not maintained here. Please solve this issue with upstream (https://github.com/catchorg/Catch2).

I have opened a PR there: https://github.com/catchorg/Catch2/pull/2619
No idea how long it gonna take to merge it and then have it updated here after that.

joto:
It has been merged upstream, but note that Osmium still uses Catch v2, not the current v3, so unless somebody ports you change back to v2, this might not end up there. I am closing this issue here now because there is nothing to be done here. I'll update the enclosed Catch version before releasing a new Osmium version to whatever is current at that time (but still Catch v2), I can't do more at the moment.

barracuda156:
Thank you. Do you suggest backporting with upstream to get it into catch2?

PR #368, opened 2023-11-27
Implement and test file seek, extract common "read_exactly" for reuse

BenWiederhake:
This PR:
- Extracts the private method `osmium::io::detail::PbfParser::read_exactly` as a generic, easily reusable function
- Implements `osmium::util::file_seek` for random access in files (currently unused, see below)
- Tests the seek feature (which also uses `read_exactly`)

This works toward #367, which is where `file_seek` will be very useful.

If you think `file_seek` should instead live in `experimental`, I'd be happy to move the function there.

### DIFF
--- a/include/osmium/util/file.hpp
+++ b/include/osmium/util/file.hpp
@@ -223,6 +223,22 @@ namespace osmium {
         }
 
         /**
+         * Set current offset into file.
+         *
+         * @param fd Open file descriptor.
+         * @param offset Desired absolute offset into the file
+         */
+        inline void file_seek(int fd, size_t offset) noexcept {
+#ifdef _MSC_VER
+            osmium::detail::disable_invalid_parameter_handler diph;
+            // https://msdn.microsoft.com/en-us/library/1yee101t.aspx
+            _lseeki64(fd, static_cast<__int64>(offset), SEEK_SET);
+#else
+            ::lseek(fd, offset, SEEK_SET);
+#endif
+        }
+
+        /**
          * Check whether the file descriptor refers to a TTY.
          *
          * @param fd Open file descriptor.
--- a/test/t/io/test_file_seek.cpp
+++ b/test/t/io/test_file_seek.cpp
@@ -1,0 +1,61 @@
+#include "catch.hpp"
+
+#include "utils.hpp"
+
+#include <osmium/io/detail/read_write.hpp>
+#include <osmium/util/file.hpp>
+
+/**
+ * Can read and seek around in files.
+ */
+TEST_CASE("Seek and read in files") {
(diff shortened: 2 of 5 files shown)


BenWiederhake:
Changes since last push:
- Added `#include <cassert>` to read_write.hpp, since I use assert(). Found through iwyu, even though its output is incredibly noisy.

include/osmium/util/file.hpp
```diff
@@ -222,6 +222,22 @@ namespace osmium {
...
+         * Set current offset into file.
+         *
+         * @param fd Open file descriptor.
+         * @param offset Desired absolute offset into the file
+         */
+        inline void file_seek(int fd, size_t offset) noexcept {
+#ifdef _MSC_VER
+            osmium::detail::disable_invalid_parameter_handler diph;
+            // https://msdn.microsoft.com/en-us/library/1yee101t.aspx
+            _lseeki64(fd, offset, SEEK_SET);
```
joto, line 235: We'll need a cast here (and below for the `lseek`) because `off_t` is a signed type and `size_t` is unsigned, to avoid warnings.

test/t/io/test_file_seek.cpp
```diff
@@ -0,0 +1,62 @@
...
+    };
+    const seek_expectation expectations[] = {
+        { 0x00, {0x1f, 0x8b, 0x08, 0x08, 0x19, 0x4a, 0x18, 0x54} },
+        { 0x00, {0x1f, 0x8b, 0x08, 0x08, 0x19, 0x4a, 0x18, 0x54} }, /* repeat / jump back */
+        { 0x21, {0x56, 0xc6, 0x18, 0xc3, 0xea, 0x6d, 0x4f, 0xe0} }, /* unaligned */
+        { 0xb3, {0xcd, 0x0a, 0xe7, 0x8f, 0xde, 0x00, 0x00, 0x00} }, /* close to end */
+        { 0x21, {0x56, 0xc6, 0x18, 0xc3, 0xea, 0x6d, 0x4f, 0xe0} }, /* "long" backward jump */
+    };
+    for (const auto& expect : expectations) {
+        printf("Trying to seek to %lu\n", expect.offset);
```
joto, line 27: Please remove this line, tests should usually not produce output.

joto (changes requested):
Looks good, except the two small things I wrote down.

I am getting some weird effects here that github actions seems not to be testing your newsest commit. Hopefully this sorts itself out when you add those changes.

BenWiederhake:
Changes since last push:
- Added the cast to __int64, that seems to be the type that's also used in other MSVC-specific implementations, like resize_file.
- Removed the print from the test

PR #375, opened 2024-06-09
replace dead site by official forum

matkoniecz:
https://help.openstreetmap.org/ has "NOTICE: help.openstreetmap.org is no longer in use from 1st March 2024. Please use the OpenStreetMap Community Forum"


matkoniecz:
If that PR would be accepted I can open similar ones for other repos.

PR #380, opened 2024-11-04
Fix LTO linker warning

frankbielig:
`warning: ‘__builtin_strlen’ reading 1 or more bytes from a region of size 0 [-Wstringop-overread]`

### DIFF
--- a/include/osmium/builder/osm_object_builder.hpp
+++ b/include/osmium/builder/osm_object_builder.hpp
@@ -470,7 +470,7 @@ namespace osmium {
              * @pre @code strlen(user) < 2^16 - 1 @endcode
              */
             TDerived& set_user(const char* user) {
-                const auto len = std::strlen(user);
+                const auto len = (user == nullptr) ? 0 : std::strlen(user);
                 assert(len < std::numeric_limits<string_size_type>::max());
                 return set_user(user, static_cast<string_size_type>(len));
             }


joto:
Can you explain why this gets rid of the warning? It seems to me the reason for the warning is something else and this "fix" only confuses the compiler enough that it doesn't print the warning any more. Problem is that this doesn't really save us, there are other cases where this warning shows up. So we still have to disable that warning.

There shouldn't be cases where `set_user()` is called with a `nullptr`.

frankbielig:
It is really the only place I get this warning. It is triggered by 

```c++
::osmium::memory::Buffer buffer(10000);
::osmium::builder::RelationBuilder relationBuilder(buffer);
::osmium::Relation&  r = relationBuilder.object();

relationBuilder.set_user(relation.user());`
```

The variable `relation` is the parameter of a `::osmium::handler::Handler::relation()` callback. Do I have to check the user against `nullptr` at this point?

joto:
> It is really the only place I get this warning.

That might be true in your code, but other code has similar problems in different places.

> relationBuilder.set_user(relation.user());

Unless you are doing something strange `relation.user()` will never return a `nullptr`! So your check should never be necessary. The reason for the warning is different, it has something to do with the way we allocate memory and instantiate OSM objects in our buffer and then add additional information like the user name after the instantiated objects.

I believe your change merely seems to make the code sufficiently more complex that the compiler can't be sure any more and doesn't produce the warning. If you compile the libosmium tests you will see a different place showing the same problem where your fix doesn't work. We need a better way to fix this problem. :-(

frankbielig:
Ok. Then I will withdraw my PR.

PR #382, opened 2024-12-09
Use 64-bit integers for OSM relation IDs in RelationMap classes.

### LINKED ISSUE #381: Int32 for OSM relation IDs in RelationsMapStash and RelationsMapIndex causes overflow => osmium tool works incorrectly
## What version of libosmium are you using?

Latest release

## What did you do exactly?

I found the issue while running `osmium-extract` command. My data had 3 relations: 2 relation with nodes members, and one "super-relation" (ID r4350609883), which has the previous two relations as members.

## What did you expect to happen?

I expected to have all the three relations after the `osmium-extract` command.

## What did happen instead?

My "super-relation" was not in the output data after `osmium-extract` command.

## What did you do to try analyzing the problem?

The problem was caused by the `osmium-tool`. It uses `libosmium` for the `osmium-extract` command, e.g. in the code for osmium-extract with SMART strategy [here](https://github.com/osmcode/osmium-tool/blob/master/src/extract/strategy_smart.cpp) it uses classes `osmium::index::RelationsMapXXX` from `libosmium`.
The RelationMapXXX classes have fields with maps, where the key and value type is `uint32_t`.
Specifically, [here for RelationsMapIndex](https://github.com/osmcode/libosmium/blob/master/include/osmium/index/relations_map.hpp#L174) and [here for RelationsMapStash](https://github.com/osmcode/libosmium/blob/master/include/osmium/index/relations_map.hpp#L315).

As we know, OSM IDs are 64bit. In case relation IDs are bigger than max uint32, the values silently overflow, and we have some "parent_id" values for relations which differ from actual relation IDs.

That is how I lost my "super-relation" with ID `4350609883`, which suddenly transformed into `55642587` (overflow: 4294967295-4350609883+1) for the parent relation.

alexeisolodovnicov-tomtom:
fixes #381

### DIFF
--- a/include/osmium/index/relations_map.hpp
+++ b/include/osmium/index/relations_map.hpp
@@ -171,8 +171,8 @@ namespace osmium {
             friend class RelationsMapStash;
             friend class RelationsMapIndexes;
 
-            using map_type = detail::flat_map<osmium::unsigned_object_id_type, uint32_t,
-                                              osmium::unsigned_object_id_type, uint32_t>;
+            using map_type = detail::flat_map<osmium::unsigned_object_id_type, uint64_t,
+                                              osmium::unsigned_object_id_type, uint64_t>;
 
             map_type m_map;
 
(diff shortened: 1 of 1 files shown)


joto:
As pnorman alludes to this would double the memory usage for everybody, even those who do do not have 64 bit relation IDs, so this isn't a viable solution. You'll need something more complex than this.

PR #392, opened 2025-08-15
Only start padding with zeros at 10^n

ginnyTheCat:
This makes sure 10 nodes still only have 1 digit as we are going from 0-9. The same applies to 100 or 1000 etc nodes ofc. Not a huge change, just a small polish.

As `way.nodes().size()`, `relation.members().size()` and `changeset.num_comments()` are all unsigned subtracting one should be ok even if they are zero.

Before:
node counts prefixed with 0

After:
node counts not prefixed with 0


joto:
Now the functionality is complex enough that it should go into its own function and get a unit test.

joto:
And maybe we should just use `snprintf` to tell us how long the output will be? See second paragraph under [Notes](https://en.cppreference.com/w/cpp/io/c/snprintf#Return_value). Unless that's much more expensive than the `log()`.

PR #394, opened 2025-08-24
Add bit to distinguish between tagged and untagged nodes [libosmium part]

hannahbast:


hannahbast:
@joto Thanks for the comment + sorry for the spam. We need this bit for our `osm2rdf` tool, see https://github.com/ad-freiburg/osm2rdf/pull/120 . I forked https://github.com/osmcode/libosmium for that, and accidentally opened a PR here (and immediately closed it again after I realized my mistake). Also, the current PR has a small bug, which, however, we have fixed in the meantime.

For us, the bit that says whether a node is tagged or untagged is quite important and useful. I wonder: Why would you find it useful?

PR #398, opened 2025-10-21
Report progress when reading PBF files

victor-pavlychko:
Looks like the `m_offset_ptr` value is never updated when using `PBFParser` with `DummyDecompressor`.


victor-pavlychko:
Thanks for merging!

